PR #18, opened 2015-04-11
Simplify script for checking tarball + fix validation error

Rufflewind:
- Make the script for checking the source tarball simpler by removing the `if` test (unless there's a reason I didn't anticipate?).
  - Also add `language: c` to please the [validator](http://yaml.travis-ci.org).
  - Can just use `cabal install dist/*-*.tar.gz` if there is no source of confusion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -99,6 +99,6 @@ script:
 
  # check that the generated source-distribution can be built & installed
  - SRC_TGZ=$(cabal info . | awk '{print $2;exit}').tar.gz &&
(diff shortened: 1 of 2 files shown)


hvr:
@Rufflewind I take it you tested those changes?

Rufflewind:
I tested them locally, and again on Travis, so it seems to work.

I made another change to run the `cd dist` under a subshell to avoid affecting later commands.

hvr:
thx

PR #33, opened 2015-09-20
Add `cabal haddock` to the test script

goldfirere:
I just released a package that couldn't haddock. I successfully added `cabal haddock` to my Travis settings, and thought others might want to as well.


hvr:
subsumed by #85

PR #38, opened 2015-11-20
Warn about GHC==x.y format in tested-with

phadej:
The proper solution would be to use later form in cabal files, but...

### DIFF
--- a/make_travis_yml.hs
+++ b/make_travis_yml.hs
@@ -3,7 +3,9 @@
 -- NB: This code deliberately avoids relying on non-standard packages
 
 import Control.Monad
+import Data.Maybe (mapMaybe)
 import Data.List
+import Data.Version (showVersion)
 import System.Environment
 import System.Exit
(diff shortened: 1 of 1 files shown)


hvr:
This is semantically inaccurate; `GHC==x.y` is plain wrong; `tested-width` follows the very same rules as `build-depends`

hvr:
I'd rather accept a patch that tells the user how to fix his `.cabal` file properly than support this malpractice

phadej:
Makes sense. Will do.

phadej:
@hvr

```
*WARNING* 'tested-with:' uses two digit GHC versions: 7.6, 7.8, 7.10
*INFO* Use wild-card format, for example 'tested-with: GHC ==7.10.*'
*ERROR* no known GHC version is allowed by the 'tested-with' specification
```

phadej:
AFAICS this is done in v2 script.

PR #46, opened 2016-05-21
Add instructions for `make_travis_yml.hs` script

Gabriel439:
I found the `make_travis_yml.hs` script in this repository extremely useful but I only learned how to use the script from another satisfied user.  If it's not too much trouble I'd like to contribute some example instructions upstream so that people can learn how to use this script without relying on word of mouth.


hvr:
huge thanks!

PR #47, opened 2016-05-21
Fix example `make_travis_yml.hs` command

Gabriel439:
I forgot to supply the cabal file as the script argument

Sorry for the mistake!  One last change to fix the instructions


hvr:
thx!

PR #65, opened 2016-12-19
Drop obsolete 7.11 compiler from the list of known versions

peti:
The new ghc-HEAD is 8.1.x.


hvr:
Fair enough, thanks! :-)

PR #82, opened 2017-05-24
Removing dist/ in a new-build repo requires removing .ghc.environment.*, too.

peti:
If `dist/` is deleted but .ghc.environment.* is not, then the repository is in a state that's going to confuse Cabal.

Fixes https://github.com/hvr/multi-ghc-travis/issues/81.


hvr:
thanks!

PR #83, opened 2017-06-16
Add branch option

phadej:
This is useful, as I don't want to build others `master` branches. If I want to build something else, I do a PR.


hvr:
thanks!

PR #84, opened 2017-06-26
Don't generate new-test like if there are no test-suites in package d…

phadej:
…escription


phadej:
Seems this repository has `.travis.yml` but it's not enabled (and outdated as well).

- @hvr can you turn it on,
- we can then dogfood ourself :)

hvr:
@phadej it's should be enabled know... knock yourself out =)

PR #85, opened 2017-07-14
Build haddocks

phadej:
supersede #33


phadej:
Doesn't seem to work for packages without library, is the bug reported to Cabal already?

hvr:
@phadej can this be workarounded similiar to beadd4177efca8e23d2d6e9c255ecc5732662c00 ?

phadej:
@hvr it can, I just got distracted.

Also `cabal` failing with pattern-match isn't good, we have to fix that (even it exits with non-zero code then).

hvr:
Also, we may need an escape hatch like in https://github.com/ekmett/mtl/pull/43/commits/81a582b57f789401bce3dd28f6675281e8eea465

hvr:
thanks!

PR #87, opened 2017-07-21
Add sponge command

phadej:
Should we have *sponge* magic built-in, it would help updating `.travis.yml`


hvr:
I'd prefer implementing an `-o <fname>`/`--output <fname>`  flag

(nonetheless, I appreciate learning about `sponge` - I kept hacking up small 5-liner pythonscripts to implement `sponge`)

PR #88, opened 2017-07-21
Add output option

phadej:


hvr:
Awesome, thanks! This also works better with `cabal new-run` (since cabal prints logging noise to `/dev/stdout`)

PR #97, opened 2017-10-13
Add support for cabal.project files.

merijn:
make-travis-yml now accept cabal.project files as input argument in addition to
cabal files, generating a test script that properly validates the individual
packages.

### DIFF
--- a/make_travis_yml_2.hs
+++ b/make_travis_yml_2.hs
@@ -1,11 +1,7 @@
-#! /usr/bin/env runghc
--- NOTE: -XCPP + shebang require at least GHC 7.8.4; prior versions of
--- GHC don't support this. For older GHCs, remove the line and pass
--- this script manually to `runghc` (or compile it)
-
 {-# LANGUAGE Haskell2010 #-}
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE NamedFieldPuns #-}
 
 #if !defined(MIN_VERSION_Cabal)
 -- As a heuristic, if the macro isn't defined, be pessimistic and
@@ -22,23 +18,31 @@
 --     is expected to compile/work with at least GHC 7.0 through GHC 8.0
 module MakeTravisYml where
 
-import Control.Applicative ((<|>))
+import Control.Applicative ((<$>),(<$),(<*>),(<*),(*>),(<|>), pure)
 import Control.DeepSeq (force)
 import Control.Exception (evaluate)
 import Control.Monad
+import Data.Char (isAsciiLower, isAsciiUpper, isSpace)
+import qualified Data.Foldable as F
 import Data.Function
 import Data.List
 import Data.Maybe
+import Data.Monoid (Monoid(..), (<>))
+import Data.Set (Set)
+import qualified Data.Set as S
 import System.Console.GetOpt
+import System.Directory (doesDirectoryExist, getDirectoryContents)
 import System.Environment
 import System.Exit
+import System.FilePath ((</>), takeDirectory, takeFileName)
 import System.IO
 import Control.Monad.IO.Class
 import Control.Monad.Trans.Writer
 import Text.Read (readMaybe)
 
 import Distribution.Compiler (CompilerFlavor(..))
-import Distribution.Package
+import Distribution.Package hiding (Package, pkgName)
+import qualified Distribution.Package as Pkg
 import Distribution.PackageDescription (packageDescription, testedWith, package, condLibrary, condTestSuites)
 import Distribution.Text
 import Distribution.Version
@@ -49,7 +53,9 @@ import Distribution.PackageDescription (GenericPackageDescription)
 import Distribution.PackageDescription.Parse (readPackageDescription)
 import Distribution.Verbosity (Verbosity)
 #endif
-
+import Text.ParserCombinators.ReadP
+    ( ReadP, (<++), between, char, eof, munch, munch1, pfail, readP_to_S
+    , satisfy, sepBy, sepBy1, string)
 
 #if !(MIN_VERSION_Cabal(2,0,0))
 -- compat helpers for pre-2.0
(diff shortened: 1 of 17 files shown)


make_travis_yml_2.hs
```diff
@@ -1,11 +1,11 @@
-#! /usr/bin/env runghc
 -- NOTE: -XCPP + shebang require at least GHC 7.8.4; prior versions of
```
phadej: I guess we don't need comment either?

make_travis_yml_2.hs
```diff
@@ -409,15 +571,15 @@ genTravisFromCabalFile (argv,opts) fn xpkgs = runFileWriter (optOutput opts) $ d
         ]
 
     -- cabal new-test fails if there are no test-suites.
-    unless (null $ condTestSuites gpd) $ tellStrLns
+    when (hasTests cfg) $ tellStrLns
```
phadej: :+1:

make_travis_yml_2.hs
```diff
@@ -509,3 +649,253 @@ collToGhcVer cid = case simpleParse cid of
...
+expandGlob root g@(FilePathGlob FilePathRelative relGlob) = do
+    result <- expandRelGlob root relGlob
+    when (null result) $
+        putStrLnErr $ "Illegal empty package glob: " ++ prettyPathGlob g
+    return result
+
+expandRelGlob :: FilePath -> FilePathGlobRel -> IO [FilePath]
+expandRelGlob root glob0 = go glob0 ""
+  where
+    go (GlobFile glob) dir = do
```
phadej: is this globbing different from one we have for `extra-source-files`? Should we expose it from `Cabal` library (if not yet?)

merijn: This globbing code is (mostly) ripped from the globbing code in cabal-install `Distribution.Client.Glob`. I don't think Cabal has equivalent globbing code?

merijn: I think it's also more powerful/flexible than the extra-source-files globbing.

phadej:
While trying on `servant` repo I get
```
*ERROR* Parse error trying to parse: cabal.project
```


```
packages:
  servant/
  servant-client/
  servant-client-core/
  servant-docs/
  servant-foreign/
  servant-server/
  doc/tutorial/
```

phadej:
is it time to write tests?

phadej:
Edit: I added a seed test-suite, let it grow!

merijn:
Ah, I did some quick testing on the glob parsing, but I wasn't aware that globs that only specify directories were allowed. I suppose those globs in your example are treated as ending in "*.cabal" by new-build? I don't think that behaviour is documented anywhere, so that should probably be fixed in the docs too. (Should be easy enough to fix if that's what supposed to happen).

merijn:
@phadej The parser error you get is the result from a cabal.project with zero cabal files mentioned. Which was caused by me expecting globs that explicitly match cabal files. I changed the globbing to treat directory globs as ending in "*.cabal", which should solve that issue, so lemme know if the latest commit works for you.

phadej:
@merijn could you rebase?

merijn:
@phadej I fixed it to compile with the new tests, but didn't have time (I only have time sporadically until Monday) to check the test results, etc. so those still fail at the moment.

merijn:
I regenerated the .yml file for the repo which fixes the test. We should probably include some cabal.project + several dummy cabal files too for testing.

phadej:
Still doesn't work. If i change `cabal.project` to:

```
packages: servant/
  servant-client/
  servant-docs/
  servant-foreign/
  servant-server/
  doc/tutorial/
```
i.e. no empty line,
I get:

```
*ERROR* empty or missing top-level 'tested-with:' definition in .cabal file; example definition:

tested-with: GHC==8.2.1, GHC==8.0.2, GHC==7.10.3, GHC==7.8.4, GHC==7.6.3, GHC==7.4.2, GHC==7.2.2, GHC==7.0.4
```

phadej:
Ok. below helps:

```diff
--- a/make_travis_yml_2.hs
+++ b/make_travis_yml_2.hs
@@ -345,7 +345,7 @@ configFromCabalFile opts cabalFile = do
 
     when (null compilers) $ do
         putStrLnErr (unlines $
-                     [ "empty or missing top-level 'tested-with:' definition in .cabal file; example definition:"
+                     [ "empty or missing top-level 'tested-with:' definition in " ++ cabalFile ++ " file; example definition:"
                      , ""
                      , "tested-with: " ++ intercalate ", " [ "GHC==" ++ display v | v <- lastStableGhcVers ]
                      ])
```

Currently I get an union of versions, should it be an intersection? 
```
*INFO* Generating Travis-CI config for testing for GHC versions: 7.8.1 7.8.2 7.8.3 7.8.4 7.10.1 7.10.2 7.10.3 8.0.1 8.0.2 8.2.1 head
```
There are files with
```
tested-with:
  GHC==7.8.4
  GHC==7.10.3
  GHC==8.0.2
  GHC==8.2.1
```
and
```
tested-with: GHC >=7.8
```

phadej:
```
cabal: Cannot select only the dependencies (as requested by the
'--only-dependencies' flag), the packages servant-foreign-0.10.1 and
servant-0.11 are required by a dependency of one of the other targets.
```

this is probably bug in `cabal`, does @hvr knows about this?

merijn:
I went with union, since that was simplest to implement, but practically speaking we probably want something smarter than union OR intersection (i.e. detecting which component should be tested with which versions and only running those tests, rather than everything), of course that assumes that local dependencies for those components are tested with a superset of GHC versions as the components.

hvr:
@phadej afaik it's a known limitation of `--only-depedencies` that it's not clear what it shall do in a multi-package project in some cases which involve inplace/local packages being part of the dependency-set. I don't remember though if there's a cabal issue filed about it with more explaination.

phadej:
I squashed and merged this branch manually. Thanks! (also fixed the broken stuff with ghc 7.8.4)

PR #100, opened 2017-10-30
Fix cabal.project parser and improve tests.

merijn:
Parser for cabal.project files now correctly handles packages fields starting
with an empty line.

Improved golden tests to verify error/info output too.

Extended test suite with some expected failures and successes.


make_travis_yml_2.hs
```diff
@@ -71,29 +73,43 @@ versionNumbers (Version vn _) = vn
...
+    Failure err1 `mappend` Failure err2 = Failure $ err1 `mappend` err2
+    Failure err1 `mappend` Success err2 _ = Failure $ err1 `mappend` err2
+    Success err1 _ `mappend` Failure err2 = Failure $ err1 `mappend` err2
+    Success l1 o1 `mappend` Success l2 o2 =
+        Success (mappend l1 l2) (mappend o1 o2)
 
-putStrLnWarn, putStrLnInfo :: MonadIO m => String -> m ()
-putStrLnWarn m = liftIO $ hPutStrLn stderr ("*WARNING* " ++ m)
-putStrLnInfo m = liftIO $ hPutStrLn stderr ("*INFO* " ++ m)
+type YamlWriter m a = MaybeT (WriterT MakeTravisOutput m) a
```
phadej: I don't see, why `MaybeT` is used. At least in `runYamlWriter` the result is omitted. Could you add a comment on the logic?

make_travis_yml_2.hs
```diff
@@ -71,29 +73,43 @@ versionNumbers (Version vn _) = vn
...
 
--- putStrLns :: [String] -> IO ()
--- putStrLns = putStr . unlines
+putStrLnErr :: Monad m => String -> YamlWriter m a
+putStrLnErr m = do
+    lift . tell $ Failure ["*ERROR* " ++ m]
+    mzero
```
phadej: is this the reason for `MaybeT`?

merijn: Yeah, it was the simplest way I could think of to preserve the termination on errors.

merijn: I'll add a note to that effect to YamlWriter.

PR #101, opened 2017-10-30
Add a flag to filter out cabal noise from test output.

merijn:
Builds upon #100

### DIFF
--- a/make_travis_yml_2.hs
+++ b/make_travis_yml_2.hs
@@ -144,6 +144,7 @@ data Options = Options
     , optOnlyBranches :: [String]
     , optOutput :: Maybe FilePath
     , optRegenerate :: Maybe FilePath
+    , optQuietTests :: !Bool
     } deriving Show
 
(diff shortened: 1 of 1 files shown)


make_travis_yml_2.hs
```diff
@@ -71,29 +73,34 @@ versionNumbers (Version vn _) = vn
 
 #endif
 
-putStrLnErr :: MonadIO m => String -> m a
-putStrLnErr  m = liftIO $ hPutStrLn stderr ("*ERROR* " ++ m) >> exitFailure
+type YamlWriter m a = ReaderT (String -> m ()) (WriterT [String] m) a
```
phadej: Rather have a tuple to write to, for yaml and diagnostics. This is ugly.

merijn: This is part of PR #100, which I just updated to match your sense of aesthetics ;)

phadej:
Could you rebase on current master?

PR #104, opened 2017-11-24
Fix for running make_travis_yml_2.hs script under Windows

### LINKED ISSUE #103: make_travis_yml_2.hs scripts broken under Windows
I use Windows to generate the scripts. Because of using `System.FilePath.</>` the generated .travis.yml are using windows backslashes instead of unix forward slashes.

Example Travis error of Windows-generated make-travis script:

```
The command "(cd "."; cabal sdist)" exited with 0.
0.01s
$ mv "."\dist\hlogo-*.tar.gz ${DISTDIR}/
mv: cannot stat ‘.disthlogo-*.tar.gz’: No such file or directory


The command "mv "."\dist\hlogo-*.tar.gz ${DISTDIR}/" exited with 1.
```

bezirg:
Fixes #103

### DIFF
--- a/make_travis_yml_2.hs
+++ b/make_travis_yml_2.hs
@@ -34,7 +34,7 @@ import System.Console.GetOpt
 import System.Directory (doesDirectoryExist, getDirectoryContents)
 import System.Environment
 import System.Exit
-import System.FilePath ((</>), takeDirectory, takeFileName)
+import System.FilePath.Posix ((</>), takeDirectory, takeFileName)
 import System.IO
 import Control.Monad.IO.Class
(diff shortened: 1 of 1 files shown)


phadej:
Hmm, did @hvr break the master?

bezirg:
I would guess so, because my PR change is very minimal.

hvr:
@phadej yeah... I did.... as I quickfixed the script via the webui to add "8.2.2" the list of known versions w/o considering that the testsuite would be sensible  to that... :-)

phadej:
cherry-picked and merged in #107, thanks for the fix!

PR #108, opened 2017-12-21
Add ghc-8.4.1 to matrix

phadej:


.travis.yml
```diff
@@ -24,6 +24,12 @@ before_cache:
   - rm -fv $HOME/.cabal/packages/hackage.haskell.org/01-index.tar
   - rm -fv $HOME/.cabal/packages/hackage.haskell.org/01-index.tar.idx
 
+  - rm -fv $HOME/.cabal/packages/head.hackage/00-index.*
```
hvr: you should just remove the complete `$HOME/.cabal/packages/head.hackage` folder; don't bother trying to be clever; there's little benefit in caching here; and in fact there's a bug in `new-update` you'll workaround by starting with an empty package cache for `head.hackage`

PR #110, opened 2017-12-28
Add `#!/usr/bin/env runghc` to make_travis_yml_2.hs

RyanGlScott:
This allows one to invoke it as a script from a shell.


RyanGlScott:
Oops, I didn't see that issue (or https://github.com/haskell-hvr/multi-ghc-travis/pull/94, for that matter). My bad.

PR #111, opened 2017-12-31
Add shebang to the _2 version

Ongy:


hvr:
See #110 ... unfortunately doing so interacts badly with CPP

Ongy:
That's interesting. Does that depend on the the version? I just ran it with shebang on my (decently up to date arch) machine.

RyanGlScott:
@Ongy, it's a [`clang`-specific issue](https://ghc.haskell.org/trac/ghc/ticket/6132), so you're unlikely to experience problems if you're not using it.

phadej:
Non applicable anymore as this become "real" package.

PR #114, opened 2018-01-02
constrainted builds

phadej:
The example uses `servant-yaml`, yet the Haskell file there will be
iterated, but I need some experience to proceed further


phadej:
I'll try another way to do this.

PR #122, opened 2018-01-03
Generation time constraint-sets

phadej:


phadej:
Example build: https://travis-ci.org/phadej/servant-yaml/jobs/324844047

PR #126, opened 2018-01-06
Pass ${TEST} ${BENCH} to `cabal new-haddock`

RyanGlScott:
Let's mitigate the possibility that we might pick a different install-plan with `--disable-tests --disable-benchmarks`. (I haven't seen this occur in practice yet, but it wouldn't surprise me if this were possible.)


phadej:
Thanks!

PR #137, opened 2018-03-09
Fix golden test suite to accommodate for changes from 1bfa6728229c9da.

peti:
Without this change, `cabal new-test` fails.


phadej:
Thanks!

Issue #11, opened 2015-01-05
Travis supports multiple ghc's now


neothemachine:
Apparently, travis supports 7.4, 7.6 and 7.8 out of the box now: http://docs.travis-ci.com/user/languages/haskell/ Maybe this should be mentioned in the readme if only a basic setup is required.

PR #148, opened 2018-04-02
travis supports multiple versions of ghc now

juhp:
for an example see https://github.com/fedora-haskell/fedora-haskell-tools/blob/master/.travis.yml


hvr:
See also #11; there's a big documentation overhaul (read, rewrite from scratch) planned

juhp:
I think it is good to fix incorrect information in a timely manner. :)

phadej:
conflicts, not sure if the issue is addressed already.

PR #151, opened 2018-04-07
Update README to use make_travis_yml_2.hs

elben:
Fixes https://github.com/haskell-CI/haskell-ci/issues/113


elben:
Ah perhaps a lot of the content in the README that describes what the old script does?

phadej:
Yes, I think it's easier to start from scratch, than try to correct current README

phadej:
This PR is now outdated. Though README is also outdated again :(

PR #157, opened 2018-05-11
Changes 2018 05 11

phadej:


hvr:
@phadej let's do this, assuming cabal-2.2 has all the features we need for CI scripts to operate properly

phadej:
In a script we don't rely on anything in `HEAD`. Some packages might work with it better though (solver improvements, other bug fixes). For those I'll make `--cabal-install-version=head` work.

PR #159, opened 2018-06-03
Update dependencies to ShellCheck 0.5.x

peti:
The build and test suite both succeed fine with the new version.


peti:
Ping?

phadej:
Thanks, but I prefer exact version on ShellCheck (it was violating PVP previously). Changed in master

PR #161, opened 2018-06-17
Build  with installed constraints by default

phadej:


peti:
How is one supposed to generate a build that runs without the "installed" constraints? There used to be `--no-installed` to accomplish that, but that flag has been removed. Now, we do have a new `--no-unconstrained` flag instead, but that does something else; it disables the unconstrained builds that would normally run at the end. However, for use cases where "installed" constraints cannot be run (i.e. because the tool needs a newer version of `time`), this flag doesn't help because the constrained builds always fail.

PR #164, opened 2018-07-02
Omit local packages in installed constraints

phadej:
Don't merge until Ryan verifies it solves the problem.


phadej:
This worked for `mtl`, `unix` and `stm`. Good enough!

PR #174, opened 2018-09-04
Package

phadej:


hvr:
@phadej I think this one requires a rebase?

phadej:
I added `travis` command which is the default (i.e. if omitted we generate `travis` config)

PR #179, opened 2018-10-01
Don't nuke ./dist-newstyle before running new-haddock

RyanGlScott:
`rm -rf ./dist-newstyle` was a grotesque hack needed in the pre-`cabal-install-2.4` days when `Cabal` couldn't figure out that Haddocks still needed to be built after running `new-build`. Thankfully, this hack is no longer necessary as of `cabal-install-2.4` (which is the minimum version that `haskell-ci` appears to use nowadays), so we can remove it.

I tested this locally on `comonad` (a project with a custom `Setup.hs` script plus two components in its `cabal.project` file) [here](https://travis-ci.org/ekmett/comonad/jobs/435742965), and it works beautifully.


phadej:
Haddocks seems to do work in travis jobs. Thanks Ryan for a PR.

Issue #176, opened 2018-09-22
hlint is hard-coded to be installed on GHC 8.4.3 (not the latest GHC)


RyanGlScott:
https://github.com/haskell-CI/haskell-ci/blob/a9a61032ca0ca2776659a5035c6f348e43bf2657/make_travis_yml_2.hs#L812-L814

The code is hard-coding the use of `80403`. This should probably be `$LATEST_COMPILER` instead so that it stays up to date with the latest GHC releases (such as 8.6.1, for instance).

There are a couple of other lines of code in `haskell-ci` that also exhibit this problem.

phadej:
I think the job HLint ruma have to be configurable, but could be the "latest stable" in the list.

RyanGlScott:
Sure, by all means let this be a configuration option—anything else is better than how it works now :)

peti:
A solution to this issue would be very much appreciated. I recently updated my test builds from ghc 8.4.3 to 8.4.4 and was unpleasantly surprised by the fact that this means `hlint` won't be running any more.

Personally, I believe that specifying `--hlint` for a configuration that actually doesn't include ghc-8.4.3 should abort the tool with an error because no valid configuration that satisfies these options can be generated.

phadej:
Fixed in https://github.com/haskell-CI/haskell-ci/pull/204

PR #194, opened 2018-12-04
Use the latest compiler version (ghc-8.6.3) to run hlint stage

peti:
8.6.2 feels like a more intuitive choice than 8.4.3.

Related to https://github.com/haskell-CI/haskell-ci/issues/176.


RyanGlScott:
Perhaps this should be 8.6.3 now?

peti:
Good point. I force-pushed an update.

phadej:
Made fix  in #204

PR #197, opened 2018-12-17
Remove sudo:false

qrilka:
Container-based Travis envs get phased out soon. See some more details at
https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures
and
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
 language: c
-sudo: false
 
 git:
   submodules: false  # whether to recursively clone submodules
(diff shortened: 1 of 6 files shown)


hvr:
I've been already considering making it configurable since the non-container configurations are quite often more efficient for non-trivial workloads... Travis forcing our hand makes the decision a lot easier ;-)

peti:
The PR needs `cabal new-run golden  -- --accept` run to fix the expected output for the regression tests.

qrilka:
I've removed also other places with `sudo: false`

phadej:
I introduced conflicts, but merging as part of #206

PR #198, opened 2018-12-17
Don't echo the command for unconstrained builds, *run it*.

peti:
The unconstrained build phase wasn't actually executed.


phadej:
...tests...

peti:
I force-pushed a new version that fixes the tests.

phadej:
I introduced conflicts, so will fix in #206

PR #203, opened 2019-02-07
Try generic lens

phadej:


src/MakeTravisYml.hs
```diff
@@ -1606,21 +1533,26 @@ configFieldDescrs =
...
+        (set #cfgOnlyBranches)
     , PU.simpleField  "folds"
         (error "we don't pretty print")
         (sepBy parseFoldQ (munch1 isSpace))
         (\cfg -> [\_ -> cfgFolds cfg])
         (\x cfg -> cfg { cfgFolds = foldl' (flip id) (cfgFolds cfg) x })
     , PU.boolField "ghc-head"
         cfgGhcHead
-        (\b cfg -> cfg { cfgGhcHead = b })
+        (set #cfgGhcHead)
```
phadej: setting is as simple, but actually we need lenses to use parsec from `Cabal-2.4`.

Issue #129, opened 2018-01-16
Allow specifying local cabal options


RyanGlScott:
Just as there is currently a `--local-ghc-options` flag for adding `ghc-options` to the autogenerated `cabal.project` file, it would be convenient to have something similar for `flags`. I currently have to add this by hand to `free`'s `.travis.yml`, for instance:

https://github.com/ekmett/free/blob/26b5a9b947969ff85e419443ae11e26319693a00/.travis.yml#L106-L108

RyanGlScott:
Or perhaps we need a more general mechanism for adding stanzas to a `package` field in the auto-generated `cabal.project` file. (For instance, I just found myself needing to generate `library-coverage: true`.)

RyanGlScott:
I think what might be nicer is to allow users to write stub files (lousy name proposal: `cabal.project.stub`) with contents like:

```
package my-package
  library-coverage: true
```

And have `haskell-ci` append that to the end of its auto-generated `cabal.project` files. This way, we wouldn't need an abundance of flags to support every possible `Cabal` flag.

RyanGlScott:
https://github.com/haskell-CI/haskell-ci/pull/157 added constraint sets via `.haskell-ci` files. However, it's unclear to me what the relationship between that and this idea is (for instance, I'm not sure if I can use constraint sets to achieve what I want here).

phadej:
This will be resolved once next `Cabal` version is released. Hopefully soon.

RyanGlScott:
This was addressed in #129 with the addition of `raw-project` stanzas, which let you do things like this:

```
raw-project
  package my-package
    library-coverage: true
```

PR #218, opened 2019-02-11
SourceRepo

### LINKED ISSUE #184: haskell-ci breaks with source-repository-packages
I have a project ([`blank-canvas`](https://github.com/ku-fpg/blank-canvas/tree/13e3e6110282bc2ca69e05d8cb89a3c1a3e5c955)) with a `source-repository-package` in its `cabal.project` file:

```
packages: .
          ./examples

source-repository-package
  type: git
  location: https://github.com/ku-fpg/remote-monad
  tag: 3761a8ad889e9bd06a562d36cafb182513dbb6e1
```

However, `haskell-ci` does not appear to pick up on this information, as demonstrated by [this failing Travis run](https://travis-ci.org/ku-fpg/blank-canvas/jobs/440702960#L651):

```
$ cabal new-build -w ${HC} ${TEST} ${BENCH} --project-file="cabal.project" --dep -j2 all
Resolving dependencies...
cabal: Could not resolve dependencies:
[__0] trying: blank-canvas-0.7 (user goal)
[__1] next goal: remote-monad (dependency of blank-canvas)
[__1] rejecting: remote-monad-0.2, remote-monad-0.1.0.1, remote-monad-0.1.0.0
(conflict: blank-canvas => remote-monad==0.5)
[__1] fail (backjumping, conflict set: blank-canvas, remote-monad)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: blank-canvas, remote-monad
```

It seems that `haskell-ci` is generating its _own_ `cabal.project` file on each Travis run but is failing to copy over `source-repository-package` information from the original `cabal.project` file.

Unfortunately, this means that `blank-canvas` is completely unusable at the moment with `haskell-ci`.

phadej:
ping @RyanGlScott, resolves #184

this isn't perfect, but you don't need to copy `source-package-repository` manually. Tests are run, but this is `cabal` bug, I guess?

I can reliably work-around https://github.com/haskell/cabal/issues/5586 because script cannot know the names of packages referenced.

### DIFF
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -70,6 +70,11 @@ import Distribution.PackageDescription.Parse (readPackageDescription)
 import Distribution.Verbosity (Verbosity)
 #endif
 
+import qualified Distribution.FieldGrammar                    as C
+import qualified Distribution.PackageDescription.FieldGrammar as C
+import qualified Distribution.Types.SourceRepo                as C
+import qualified Text.PrettyPrint                             as PP
+
 #ifdef MIN_VERSION_ShellCheck
 import ShellCheck.Checker (checkScript)
 import qualified ShellCheck.Interface as SC
@@ -962,6 +967,11 @@ genTravisFromConfigs argv opts isCabalProject config prj@Project { prjPackages =
             OptimizationOn      -> return ()
             OptimizationOff     -> tellStrLns [ sh $ "echo 'optimization: False' >> cabal.project " ]
             OptimizationLevel l -> tellStrLns [ sh $ "echo 'optimization: " ++ show l ++ "' >> cabal.project " ]
+
+        F.forM_ (prjSourceRepos prj) $ \repo -> do
+            let repo' = PP.render $ C.prettyFieldGrammar (C.sourceRepoFieldGrammar $ C.RepoKindUnknown "unused") repo
+            tellStrLns [ sh $ "echo 'source-repository-package' >> cabal.project" ]
+            tellStrLns [ sh $ "echo '  " ++ l ++ "' >> cabal.project" | l <- lines repo' ]
 
         -- also write cabal.project.local file with
         -- @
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Sorry, I didn't see this until just now! This looks quite promising, and I'm eager to give it a try.

> Tests are run, but this is `cabal` bug, I guess?

I assume you're referring to the tests in the `source-repository-package`s themselves? If so, does [this workaround](https://github.com/haskell-CI/haskell-ci/issues/184#issuecomment-453757891) help?

> I can reliably work-around [haskell/cabal#5586](https://github.com/haskell/cabal/issues/5586) because script cannot know the names of packages referenced.

I'm not sure what you mean here. Can you be more specific?

phadej:
@RyanGlScott if I have

```
source-repository-package
  type: git
  location: https://example.com/repo.git
  tag: deadc0de
```

then just from that text information, I cannot infer what packages are there, I cannot generate

```
package example-package
  tests: false
 benchmarks: false
```

We need some extra bit of information. So for now, I'd wait until cabal bug is fixed or I get to implement some variant of #129

RyanGlScott:
Ah, I see what you mean now. I encountered that exact same issue when trying to implement my hack for this in `travis-maintenance`. I think I settled for only making it work for `locations` that end in `.../repo-name.git` (where the `.git` part is optional). It works OK for my needs, since every `source-repository-package` I use happens to look like that, but I could understand if you wish to aspire to a higher standard than that :)

I wish that `cabal.project` files gave you a way of setting options for each `source-repository-package` (similarly to how you can set options for every package with `package *`). If we had that, we could avoid the need to specify the names of each `source-repository-package` in the first place.

phadej:
Latter is difficult, given e.g. `subdirs` and someone wishing to differentiate between different packages in the source-repo-package... It's a swamp.

phadej:
(for the record, my only use case right now is also of `package-name.git` tail, so maybe i'll make a HACK anyway)

PR #223, opened 2019-02-13
Fix repository's incorrect language-classification on GitHub.com

Alhadis:
Currently, GitHub classifies `haskell-ci` as a Roff repository:

Figure 1

This stems from the abundance of `collections/cabal.project.*` files with numeric suffixes, which GitHub mistakes as man pages. This PR is a small patch to correct that [using an override](https://github.com/github/linguist/blob/master/README.md#overrides).

**NOTE:** I'm *assuming* the files contained in [`collections/`](https://github.com/haskell-CI/haskell-ci/tree/master/collections) are Haskell source. If they're not, please let me know and I'll mark them as ordinary `Text`, instead of being any particular language.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Fix incorrect language classification on GitHub.com
+collections/* linguist-vendored linguist-language=Cabal


phadej:
Technically they are Cabal configuration, so it's more like EDN or JSON than Clojure or JavaScript. I guess linguist doesn't know about Cabal files?

Alhadis:
> I guess linguist doesn't know about Cabal files?

Erm, [it actually doesn't](https://github.com/github/linguist/blob/ad0871846aa234f406421c69b769cfec256470a3/lib/linguist/languages.yml#L1986-L1997). That actually needs to be fixed upstream; I'll look into submitting a PR to have Cabal added as a language.

Since I'm (clearly) Haskell illiterate, would you be able to tell me in advance what filenames and file extensions should be associated with Cabal?

Regarding this PR: I'll mark the files as `Cabal` in advance, but also `linguist-vendored`, which will instruct Linguist not to take them into account when calculating the language stats.

Alhadis:
@phadej Alright, PR for Cabal support has been [submitted upstream](https://github.com/github/linguist/pull/4419). 👍 Once it's merged and released in a new release of Linguist, the `linguist-vendored` attribute added by this PR can be dropped.

phadej:
@Alhadis extensions are a bit tricky
- `foo.cabal` are package definition files
- and additionally we have `cabal.project`, `cabal.haskell-ci` are files which share the same syntax
    - because the only `<pkg>.cabal` file in the directory is the definition of the package
    - .... so extension can be suffix or prefix

phadej:
Yet, handling only `*.cabal` files is fine for Cabal, I guess.

phadej:
I guess this is fine for now, we can discuss Cabal + linguist in your PR.

Alhadis:
Thanks! 👍

As mentioned at `github/linguist#4419`, we don't have a way of matching prefixes, only suffixes or entire filenames. 😢 Having said that, I'd recommend adding this to `.gitattributes` too:

~~~gitattributes
cabal.haskell-ci linguist-language=Cabal
~~~

This will activate correct syntax highlighting for the file once the PR is live on GitHub.

phadej:
Fwiw, `cabal.haskell-ci` is (hopefully) become common-ish filename too, as people use this tool.

Alhadis:
I ran a search for that, but it only yielded [16 results](https://github.com/search?q=filename%3Acabal.haskell-ci+NOT+nothack&type=Code). We require languages/filetypes to have considerable distribution in-the-wild before supporting them officially (at least 100 unique repos/users).

PR #229, opened 2019-02-22
Try windows

phadej:


phadej:
Ok. Windows is not so easy. If someone have time to make it work, please do continue.

phadej:
This works, almost
- https://github.com/haskell/cabal/pull/5870
- tests in *this* package are broken, not sure why.

PR #230, opened 2019-02-22
Use explicit paths, don't add /opt/ghc/bin into PATH

phadej:


ivan-m:
As an FYI, this is now breaking some of my test-suites because they need `ghc-pkg` in the PATH.

(I can tweak the generated .travis.yml script to add this back in, but now have to remember to do this every time I re-generate it)

phadej:
Don’t expect ghc-pkg to be in PATH. Use https://hackage.haskell.org/package/ghc-pathsif you really need to get to it (in tests it’s probably fine).

PR #238, opened 2019-03-02
Print the haskell-ci version number in .travis.yml scripts

### LINKED ISSUE #237: Print the version number and commit hash for haskell-ci in generated .travis.yml files
When `haskell-ci`-generated `.travis.yml` scripts go wrong, it's sometimes difficult to tell if it's due to a bug in an old version of `haskell-ci` or not. The reason for this is that `haskell-ci`'s generated `.travis.yml` files give no indication whatsoever of what version of `haskell-ci` was used to generate it, which forces me to guess.

To make this easier, I think it would be nice if `haskell-ci` printed version information at the top of each `.travis.yml` file, similarly to what `hpack` does for its generated `.cabal` files:

```
-- This file has been generated from package.yaml by hpack version 0.31.0.
--
-- see: https://github.com/sol/hpack
--
-- hash: cbcda7d66dae228545cd78a42c4c4edf8ec2a746587f3aac59ba8a8775252cc5
```

RyanGlScott:
Fixes #237.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
+# version: 0.1.0.0
+#
 language: c
 dist: xenial
 
--- a/fixtures/cabal.project.copy-fields.all.travis.yml
+++ b/fixtures/cabal.project.copy-fields.all.travis.yml
@@ -4,6 +4,8 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
+# version: 0.1.0.0
+#
 language: c
 dist: xenial
 
--- a/fixtures/cabal.project.copy-fields.none.travis.yml
+++ b/fixtures/cabal.project.copy-fields.none.travis.yml
@@ -4,6 +4,8 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
+# version: 0.1.0.0
+#
 language: c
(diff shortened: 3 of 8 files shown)


RyanGlScott:
I'm not sure I understand what is going wrong with the Travis build:

```
$ (cd haskell-ci-* && doctest --fast -XDeriveFoldable -XDeriveFunctor -XDeriveGeneric -XDeriveTraversable src)
Loaded package environment from /tmp/dist-test.1rUS/.ghc.environment.x86_64-linux-8.6.3

src/HaskellCI.hs:115:1: error:
    Could not load module ‘Paths_haskell_ci’
    it is a hidden module in the package ‘haskell-ci-0.1.0.0’
    Use -v to see a list of the files searched for.
    |
115 | import Paths_haskell_ci (version)
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
The command "(cd haskell-ci-* && doctest --fast -XDeriveFoldable -XDeriveFunctor -XDeriveGeneric -XDeriveTraversable src)" exited with 1.
```

phadej:
The ad-hoc `doctest` running doesn't work well with autogenerated files :(

RyanGlScott:
Crazy idea: perhaps we should define a `DOCTEST` CPP macro and guard the relevant pieces of code with `#ifndef DOCTEST ... #endif`?

phadej:
I found another way in #239

PR #243, opened 2019-03-05
Document copy-fields, and make local-ghc-options always fire

RyanGlScott:
I attempted to use the `copy-fields` feature the other day only to realize that I have no idea how it is intended to work. This is my attempt at adding documentation for it.

### DIFF
--- a/cabal.haskell-ci
+++ b/cabal.haskell-ci
@@ -61,6 +61,21 @@ constraint-set deepseq-1.4
   ghc: (>= 7.8 && <7.10) || == 8.2.2
   constraints: deepseq ==1.4.*
 
+-- Copy over additional properties specified in a cabal.project file.
+-- Possible values are:
+--   none: Do not copy any properties from the cabal.project file.
+--   some: Copy a subset of the properties from the cabal.project file. (Default)
+--         The fields and sections that are copied over are:
+--
+--         * constraints
+--         * allow-newer
+--         * reorder-goals
+--         * max-backjumps
+--         * optimization
+--         * source-repository-package
+--   all: Copy every property from the cabal.project file.
+copy-fields: some
+
 -- Include these fields "as is" in generated cabal.project
 raw-project
   keep-going: False
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -822,12 +822,6 @@ genTravisFromConfigs argv opts isCabalProject config prj@Project { prjPackages =
                     tellStrLns
                         [ sh $ "echo 'allow-newer: " ++ s ++ "' >> cabal.project"
                         ]
-                unless (null (cfgLocalGhcOptions config)) $ forM_ pkgs $ \Pkg{pkgName} -> do
-                    let s = unwords $ map (show . PU.showToken) $ cfgLocalGhcOptions config
-                    tellStrLns
-                        [ sh $ "echo 'package " ++ pkgName ++ "' >> cabal.project"
-                        , sh $ "echo '  ghc-options: " ++ s ++ "' >> cabal.project"
-                        ]
 
                 when (prjReorderGoals prj) $
                     tellStrLns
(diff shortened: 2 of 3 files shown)


cabal.haskell-ci
```diff
@@ -61,6 +61,13 @@ constraint-set deepseq-1.4
   ghc: (>= 7.8 && <7.10) || == 8.2.2
   constraints: deepseq ==1.4.*
 
+-- Copy over additional properties specified in a cabal.project file.
+-- Possible values are:
+--   none: Do not copy any properties from the cabal.project file.
+--   some: Copy particular properties (???) from the cabal.project file. (Default)
```
RyanGlScott: It's rather unclear to me which properties `some` copies over. Feedback on the documentation for this would be appreciated.

RyanGlScott:
Another baffling thing about `copy-fields` is that it apparently interacts with `local-ghc-options`. For instance, if you specify `copy-fields: some`, then any `local-ghc-options` gets copied to the generated `.travis.yml` file, but if you specify `copy-fields: all`, then `local-ghc-options` is ignored! I can't tell if this is intended behavior or not, so if it is, we should document this as well.

RyanGlScott:
In fact, looking at the `haskell-ci` source code, it appears that there are _several_ additional things that `copy-fields: some` does that `copy-fields: all` does not. Here is the entirety of the work that `copy-fields: all` does:

https://github.com/haskell-CI/haskell-ci/blob/45b6b6874130b5d3502da21ea2e295ecb30efc28/src/HaskellCI.hs#L809-L813

Contrast this with the many things that `copy-fields: some` does!

https://github.com/haskell-CI/haskell-ci/blob/45b6b6874130b5d3502da21ea2e295ecb30efc28/src/HaskellCI.hs#L814-L850

phadej:
the `local-ghc-options` behavior is accidental. I think we should generate them in both scenarios.

If you look more closely at what else `Some` path does, is it copies data from project: it reads `prj*` fields.

RyanGlScott:
> the `local-ghc-options` behavior is accidental. I think we should generate them in both scenarios.

Phew! In that case, I'll expand the scope of this PR to fix that bug as well.

RyanGlScott:
How does this look now, @phadej?

RyanGlScott:
Also, let me know if you think the title of this PR needs to change to reflect its expanded scope.

cabal.haskell-ci
```diff
@@ -61,6 +61,21 @@ constraint-set deepseq-1.4
...
   constraints: deepseq ==1.4.*
 
+-- Copy over additional properties specified in a cabal.project file.
+-- Possible values are:
+--   none: Do not copy any properties from the cabal.project file.
+--   some: Copy a subset of the properties from the cabal.project file. (Default)
+--         The lines that are copied over are:
+--
+--         * constraints
+--         * all-newer
```
phadej: allow-newer

cabal.haskell-ci
```diff
@@ -61,6 +61,21 @@ constraint-set deepseq-1.4
   ghc: (>= 7.8 && <7.10) || == 8.2.2
   constraints: deepseq ==1.4.*
 
+-- Copy over additional properties specified in a cabal.project file.
+-- Possible values are:
+--   none: Do not copy any properties from the cabal.project file.
+--   some: Copy a subset of the properties from the cabal.project file. (Default)
+--         The lines that are copied over are:
```
phadej: fields and sections

src/HaskellCI.hs
```diff
@@ -860,6 +854,13 @@ genTravisFromConfigs argv opts isCabalProject config prj@Project { prjPackages =
             , not (null l)
             ]
 
+        unless (null (cfgLocalGhcOptions config)) $ forM_ pkgs $ \Pkg{pkgName} -> do
```
phadej: I guess it's irrelevant whether `cfgRawProject` or `cfgLocalGhcOptions` comes first?

cabal.haskell-ci
RyanGlScott: Done.

cabal.haskell-ci
RyanGlScott: Done.

src/HaskellCI.hs
phadej: I actually, now. I think `rawProject` should came last, so you could do

```
local-ghc-options: -O2
raw-project:
  package tricky
    ghc-options: -O
```

i.e. you could still override "all packages" stuff in `raw-project`

RyanGlScott: I've swapped the order of `local-ghc-options` and `raw-project`.

phadej:
@RyanGlScott I released this in 0.2.1, and bumped the `master` to be 0.3

I'll try experimenting of having odd numbers be development versions. This way the version info in generated `.travis.yml` would have a little more value.

RyanGlScott:
Thanks, @phadej!

PR #244, opened 2019-03-06
Add a --version flag

RyanGlScott:
Currently, there's not a simple way to quickly ascertain the `haskell-ci` version. This adds `--version`, which prints out what you would expect.

### DIFF
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -111,10 +111,6 @@ import HaskellCI.Project
 import HaskellCI.TestedWith
 import HaskellCI.Version
 
-#ifndef CURRENT_PACKAGE_VERSION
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Do you think we should have both `--version` and `--numeric-version`, or does just having `--version` suffice?

phadej:
`--numeric-version` would be useful for scripting, but I don't know what that script would be?

RyanGlScott:
No idea—just thought I'd be diligent and ask. (And in any case, the behavior of `haskell-ci`'s `--version` flag behaves identically to `--numeric-version`.)

phadej:
Oh, if `--version` prints just version than it's enough for scripting I guess :)

Issue #236, opened 2019-03-02
A middle ground between --tests and --no-tests


RyanGlScott:
Currently, `haskell-ci` gives you the ability to prevent building any test suites using the `--no-tests` flag. This is _almost_ what I need, but not quite. I have a project where I can't run the test suites on Travis due to the tests' [reliance on a web browser](https://github.com/ku-fpg/blank-canvas/issues/73) being run at the same time. That being said, I still want to _build_ (but not run) the test suite on Travis so as to ensure that the code doesn't bitrot. Unfortunately, `--no-tests` is too heavy-handed, as it prevents the test suites from being built in the first place.

Could there be a flag that permits the test suite to be built but not run, similarly to how benchmarks are currently handled? I'm not sure what the best name for this flag should be, or if it would make sense to make `--no-tests` further configurable to allow this.

phadej:
I was thinking of introducing `--run-tests`, `--no-run-tests` `--run-tests-job=RANGE` configuration, which is
- intersected with `--tests`
- says when to run tests.

RyanGlScott:
Sounds good to me! If these flags do intersect with `--tests`, it would be good to document somewhere how that works, but otherwise I don't foresee any complications.

phadej:
Intersection means hopefully obvious: only built tests can be run, but you can run tests on less jobs. E.g.

```
haskell-ci --tests-jobs '>= 7.6' --run-tests-jobs '>=8.0' ...
```

phadej:
... and with

```
haskell-ci --no-tests --run-tests
```

you won't build (and run) any tests.

phadej:
`--run-tests` is there

PR #245, opened 2019-03-06
Fix #182 and #236

### LINKED ISSUE #182: constraint-sets don't encompass tests, benchmarks, or Haddocks

phadej:
- Add global --run-tests/--no-run-tests/--run-tests-jobs to specify
  when we actually run tests (maybe less then what we build!)

- Add steps configuration to constraint sets:
  - tests
  - run-tests
  - benchmarks
  - haddock
  - these are only booleans for now, if you need more control, make
    another constraint set!

---

cc @RyanGlScott

### DIFF
--- a/cabal.haskell-ci
+++ b/cabal.haskell-ci
@@ -73,6 +73,7 @@ osx: 8.4.4
 constraint-set deepseq-1.4
   ghc: (>= 7.8 && <7.10) || == 8.2.2
   constraints: deepseq ==1.4.*
+  constraints: binary installed
 
   -- Constraint sets accept booleans for few steps, as the main script
   -- Defaults are False.
--- a/cabal.haskell-ci
+++ b/cabal.haskell-ci
@@ -12,6 +12,19 @@ ghc-head: True
 -- remove cabal noise from test output
 -- cabal-noise: False
 
+-- Build tests. In addition to True and False you may specify
+-- a version range, e.g. >= 8.0 to build tests only in some jobs.
+tests: True
+
+-- Run tests. Note that only built tests are run. Accepts booleans or version range.
+run-tests: True
+
+-- Build benchmarks. There are no way to run benchmarks. Accepts booleans or version range.
+benchmarks: True
+
+-- Build haddocks. Accepts booleans or version range.
+haddock: True
+
 -- Run cabal check
 -- cabal-check: True
 
@@ -60,6 +73,11 @@ osx: 8.4.4
 constraint-set deepseq-1.4
   ghc: (>= 7.8 && <7.10) || == 8.2.2
   constraints: deepseq ==1.4.*
+
+  -- Constraint sets accept booleans for few steps, as the main script
+  -- Defaults are False.
+  -- These fields don't accept version ranges: you should rather create
+  -- another constraint set.
 
   -- tests: False
   -- run-tests: False
--- a/cabal.haskell-ci
+++ b/cabal.haskell-ci
@@ -61,6 +61,11 @@ constraint-set deepseq-1.4
   ghc: (>= 7.8 && <7.10) || == 8.2.2
   constraints: deepseq ==1.4.*
 
+  -- tests: False
+  -- run-tests: False
+  -- benchmarks: False
+  haddock: True
+
 -- Copy over additional properties specified in a cabal.project file.
 -- Possible values are:
 --   none: Do not copy any properties from the cabal.project file.
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -964,7 +964,7 @@ doctestJobVersionRange = orLaterVersion $ mkVersion [8,0]
 -- /Note:/ same argument work for hlint too, but not exactly
 --
 doctestArgs :: GenericPackageDescription -> [[String]]
-doctestArgs gpd =
+doctestArgs gpd = nub $
     [ libraryModuleArgs c
     | c <- flattenPackageDescription gpd ^.. L.library . traverse
     ] ++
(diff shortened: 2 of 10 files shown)


src/HaskellCI.hs
```diff
@@ -761,18 +743,24 @@ genTravisFromConfigs argv opts isCabalProject config prj@Project { prjPackages =
...
+            let shForCs = shForJob versions' (csGhcVersions cs)
+            let constraintFlags = concatMap (\x ->  " --constraint='" ++ x ++ "'") (csConstraints cs)
+
             foldedTellStrLns' FoldConstraintSets name ("Constraint set " ++ name) folds $ tellStrLns
-                [ shForJob versions' (csGhcVersions cs) $
-                  cmd ++ " " ++ constraintFlags ++ " all"
-                , ""
+                [ shForCs $ "CSTEST=" ++ if csTests cs then "--enable-tests" else "--disable-tests"
+                , shForCs $ "CSBENCH=" ++ if csBenchmarks cs then "--enable-benchmarks" else "--disable-benchmarks"
+                , shForCs $ "${CABAL} v2-build -w ${HC} ${CSTEST} ${CSBENCH} all " ++ constraintFlags
```
RyanGlScott: This is a really minor thing, but is there any reason to use `v2-` here over `new-`, like the rest of `haskell-ci` uses?

cabal.haskell-ci
```diff
@@ -61,6 +61,11 @@ constraint-set deepseq-1.4
   ghc: (>= 7.8 && <7.10) || == 8.2.2
   constraints: deepseq ==1.4.*
 
+  -- tests: False
+  -- run-tests: False
+  -- benchmarks: False
+  haddock: True
```
RyanGlScott: It would be nice to document `run-tests` here.

src/HaskellCI.hs
```diff
@@ -761,18 +743,24 @@ genTravisFromConfigs argv opts isCabalProject config prj@Project { prjPackages =
...
+
+            let shForCs = shForJob versions' (csGhcVersions cs)
+            let constraintFlags = concatMap (\x ->  " --constraint='" ++ x ++ "'") (csConstraints cs)
+
             foldedTellStrLns' FoldConstraintSets name ("Constraint set " ++ name) folds $ tellStrLns
-                [ shForJob versions' (csGhcVersions cs) $
-                  cmd ++ " " ++ constraintFlags ++ " all"
-                , ""
+                [ shForCs $ "CSTEST=" ++ if csTests cs then "--enable-tests" else "--disable-tests"
+                , shForCs $ "CSBENCH=" ++ if csBenchmarks cs then "--enable-benchmarks" else "--disable-benchmarks"
```
RyanGlScott: Since you factor our `CSTEST` and `CSBENCH`, would it make sense to also factor out `constraintFlags` into its own environment variable?

src/HaskellCI.hs
phadej: I'll slowly convert everything to `v2-` at some point.

src/HaskellCI.hs
phadej: I did the opposite :)

src/HaskellCI.hs
```diff
@@ -761,18 +743,25 @@ genTravisFromConfigs argv opts isCabalProject config prj@Project { prjPackages =
...
             ]
         forM_ constraintSets $ \cs -> do
             let name = csName cs
-            let constraintFlags = concatMap (\x ->  " --constraint='" ++ x ++ "'") (csConstraints cs)
-            let cmd | csRunTests cs = "${CABAL} new-test  -w ${HC} --enable-tests  --enable-benchmarks"
-                    | otherwise     = "${CABAL} new-build -w ${HC} --disable-tests --disable-benchmarks"
             tellStrLns [ comment $ "Constraint set " ++ name ]
+
+            let shForCs = shForJob versions' (csGhcVersions cs)
+            let testFlag = if csTests cs then " --enable-tests" else " --disable-tests"
```
RyanGlScott: The space character in `" --enable-tests"` is unnecessary, since this ends up being fed into `unwords`, which adds the necessary spacing anyways. (Similarly for `" --disable-tests"`, `" --enable-benchmarks"`, and `" --disable-benchmarks"`.)

phadej: good catch.

phadej: it's interesting that per package `tests: False` overrides cli argument `--enable-tests` btw. I find that slightly weird.

phadej:
I also pushed (hopefully) a fix for #246, if you, @RyanGlScott, mind to check as you are testing this branch

src/HaskellCI.hs
```diff
@@ -1010,7 +1010,7 @@ hlintJobVersionRange vs HLintJobLatest = case S.maxView vs of
 hlintJobVersionRange _ (HLintJob v)   = thisVersion v
 
 hlintArgs :: GenericPackageDescription -> [[String]]
-hlintArgs gpd =
+hlintArgs gpd = nub $
```
phadej: note, because extensions could be different let us not `concat` those.

RyanGlScott:
I just tried it out on `lens` and `rcu`, and it appears to do the right thing.

PR #251, opened 2019-03-20
Change how HC and HCPKG is calculated

### LINKED ISSUE #249: Travis can't find PPA-installed GHC during the 'install' phase
See for example here: https://travis-ci.org/haskell/hackage-security/jobs/509084336#L313 (corresponding PR: https://github.com/haskell/hackage-security/pull/220):

```
$ ghc-head --version
ghc-head: command not found
[...]
/home/travis/.travis/functions: line 106: /opt/ghc-pkg/bin/ghc-head: No such file or directory
```

Despite the fact that GHC was succesfully installed from the PPA, during the `install` phase Travis for some reason thinks that it's not present on the system. This causes the step that adds `allow-newer` lines to `~/.cabal/config` fail, causing the GHC HEAD configuration to fail as well.

phadej:
Fixes #249

@23Skidoo  can you check this; this should work better (`cabal` didn't like `-w ghc-head` at some point)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,8 +54,8 @@ matrix:
     - compiler: "ghc-head"
 
 before_install:
-  - HC=/opt/ghc/bin/${CC}
-  - HCPKG=${HC/ghc/ghc-pkg}
+  - HC=$(echo "/opt/$CC/bin/ghc" | sed 's/-/\//')
+  - HCPKG="$HC-pkg"
   - unset CC
   - CABAL=/opt/ghc/bin/cabal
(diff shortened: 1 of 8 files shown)


phadej:
@23Skidoo also note that `ghc-8.7.20181217` i.e. old, and `head.hackage` went forward, so some patches might "break" things (e.g. `cabal-doctest` doesn't work). Hopefully not an issue for `hackage-security`.

23Skidoo:
@phadej This patch seems to work.

PR #262, opened 2019-04-01
Implement a --patches flag

### LINKED ISSUE #257: Allow specifying patch files
99% of the time, `haskell-ci` will just work for your project out of the box. The remaining 1% of your projects, however, often require a handful of manual tweaks in order to make everything work just right. An example of this can be found in the `base-compat` library, which adds some additional goodies to `haskell-ci`'s autogenerated `.travis.yml` file in the form of [this patch](https://github.com/haskell-compat/base-compat/blob/257e6971d5d69720388f6cad5eac58c937176a48/travis.yml.patch).

To make this process a little more streamlined, it would be convenient if `haskell-ci` could automatically apply these patch files as a part of its `.travis.yml` generation process. Of course, there is potential for fragility here, since it's possible that a patch file created at one point in time will no longer apply with a later version of `.travis.yml`, but this would at the very least substantially simplify the process of actually _using_ the patch in the event that one has to tweak the original, unpatched `.travis.yml` file.

RyanGlScott:
This allows specifying one or more `.patch` files with a new `--patches` flag. The actual patching occurs in a temporary file in case the user did not specify `--output`.

Fixes #257.


RyanGlScott:
Presumably, we'll want to add tests for this feature. What would be the best way to accomplish this?

src/HaskellCI/Config.hs
```diff
@@ -185,6 +187,8 @@ configGrammar = Config
         ^^^ metahelp "JOB" "Jobs to additionally build with OSX"
     <*> C.monoidalFieldAla    "apt"                       (alaSet' C.NoCommaFSep C.Token')    #cfgApt
         ^^^ metahelp "PKG" "Additional apt packages to install"
+    <*> C.monoidalFieldAla    "patches"                   (C.alaList' C.FSep C.Token')        #cfgPatches
```
phadej: I'd use `NoCommaFSep` here too. Allowing commas is redundant.

phadej: And let's name the field `travis-patches`

src/HaskellCI.hs
```diff
@@ -208,6 +212,36 @@ genTravisFromConfigs argv config prj versions' = do
...
+-- | Adjust the generated Travis YAML output with patch files, if specified.
+-- We do this in a temporary file in case the user did not pass --output (as
+-- it would be awkward to patch the generated output otherwise).
+patchTravis
+    :: (MonadIO m, MonadMask m)
+    => Config -> [String] -> m [String]
+patchTravis cfg ls
+  | null patches = pure ls
+  | otherwise =
+      withSystemTempFile ".travis.yml.tmp" $ \fp h -> liftIO $ do
```
phadej: what does `patch` do if patch application fails; it doesn't leave any conflict/orig/ etc files, does it?

src/HaskellCI/Config.hs
RyanGlScott: Done and done.

src/HaskellCI.hs
RyanGlScott: I haven't given much thought to how we should handle errors yet (which is one of the main reasons why I marked this as a WIP).

One option is to simply let the `patch` tool error out and cause `haskell-ci` to terminate. This could potentially leave some conflict files, but then again, it would leave them in the system's temporary directory, so perhaps this isn't so bad. Alternatively, we could try some more robust error-handling scheme.

What would you prefer?

haskell-ci.cabal
```diff
@@ -138,17 +138,20 @@ library haskell-ci-internal
     , mtl           >=2.2.2
     , parsec        ^>=3.1.13.0
     , pretty        ^>=1.1.3.3
+    , process       >=1.4.3.0
```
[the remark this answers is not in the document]
phadej: There aren't GHC with `process-1.5`:

```
[FL973] ~/mess/composition-1.0.2.1 % ghc-pkg-8.0.2 list | grep process
    process-1.4.3.0
[FL973] ~/mess/composition-1.0.2.1 % ghc-pkg-8.2.2 list | grep process
    process-1.6.1.0
```

src/HaskellCI.hs
phadej: You would be the user, whatever is fine with you :)

haskell-ci.cabal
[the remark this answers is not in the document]
RyanGlScott: True, but there is a `process-1.5.0.0` release on Hackage, so it seems strange to specifically exclude that from being chosen from build plans.

src/HaskellCI.hs
RyanGlScott: Hah, fair enough. I'll do some experimentation to see which choice seems better from an ergonomic perspective.

haskell-ci.cabal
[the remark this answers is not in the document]
phadej: we don't test it, so we exclude it.

src/HaskellCI.hs
phadej: Any outcome?

RyanGlScott: Sorry for the delay—I haven't had a chance to work on this in a while.

Feel free to close this if you wish, as I can always reopen later if need be.

phadej: No, it can be open, if you plan to continue to work on this

RyanGlScott: Alright, I've had a chance to characterize the different ways this can fail, and what error messages you currently get in each situation:

1. The `.patch` file is missing:

   ```
   $ ../dist-newstyle/build/x86_64-linux/ghc-8.6.4/haskell-ci-0.3.20190413/x/haskell-ci/build/haskell-ci/haskell-ci regenerate
   *INFO* Generating Travis-CI config for testing for GHC versions: 7.8.1 7.8.2 7.8.3 7.8.4 7.10.1 7.10.2 7.10.3 8.0.1 8.0.2 8.2.1 8.2.2 8.4.1 8.4.2 8.4.3 8.4.4 8.6.1 8.6.2 8.6.3 8.6.4 8.6.5 8.8.1
   *ERROR* Cannot find cabal.project.travis-patch.patch
   ```
2. The `.patch` file contains garbage:

   ```
   $ ../dist-newstyle/build/x86_64-linux/ghc-8.6.4/haskell-ci-0.3.20190413/x/haskell-ci/build/haskell-ci/haskell-ci regenerate
   *INFO* Generating Travis-CI config for testing for GHC versions: 7.8.1 7.8.2 7.8.3 7.8.4 7.10.1 7.10.2 7.10.3 8.0.1 8.0.2 8.2.1 8.2.2 8.4.1 8.4.2 8.4.3 8.4.4 8.6.1 8.6.2 8.6.3 8.6.4 8.6.5 8.8.1
   patch: **** Only garbage was found in the patch input.
   haskell-ci: callProcess: patch "-i" "cabal.project.travis-patch.patch" "-s" "/tmp/.travis.yml3806-0.tmp" (exit 2): failed
   ```
3. The `.patch` file doesn't apply cleanly:

   ```
   $ ../dist-newstyle/build/x86_64-linux/ghc-8.6.4/haskell-ci-0.3.20190413/x/haskell-ci/build/haskell-ci/haskell-ci regenerate
   *INFO* Generating Travis-CI config for testing for GHC versions: 7.8.1 7.8.2 7.8.3 7.8.4 7.10.1 7.10.2 7.10.3 8.0.1 8.0.2 8.2.1 8.2.2 8.4.1 8.4.2 8.4.3 8.4.4 8.6.1 8.6.2 8.6.3 8.6.4 8.6.5 8.8.1
   1 out of 1 hunk FAILED -- saving rejects to file /tmp/.travis.yml4518-0.tmp.rej
   haskell-ci: callProcess: patch "-i" "cabal.project.travis-patch.patch" "-s" "/tmp/.travis.yml4518-0.tmp" (exit 1): failed
   ```

Some observations:

* Only (1) gives you an `*ERROR*`-prefixed message, due to the fact that I check for the absence of `.patch` files directly in `haskell-ci`. The other error messages simply barf out what `patch` tells you, so they're not formatted quite as nicely.
* Currently, all three forms of errors give you exit code 1, which is nice. However, the fact that (2) and (3) give you exit code 1 is probably just due to the fact that `patch` happens to return exit code 1, so I suppose there's nothing that guarantees this.

I don't know how to check for (2) or (3) directly in `haskell-ci`, so it seems likely that we'll have to rely on `patch` to error out. That being said, we could intercept the `patch` subprocess, check for its exit code, and format the errors a little nicer in the event that we receive a non-zero exit code. Do you think that would be worth it?

haskell-ci.cabal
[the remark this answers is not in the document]
RyanGlScott: Very well, I've applied this change.

src/HaskellCI.hs
RyanGlScott: I've just pushed a change so that we now use `readProcessWithExitCode` instead of `callProcess`. This means that we now have the following for scenarios (2) and (3):

2. ```
   $ ../dist-newstyle/build/x86_64-linux/ghc-8.6.4/haskell-ci-0.3.20190413/x/haskell-ci/build/haskell-ci/haskell-ci regenerate -o cabal.project.travis-patch.travis.yml 
   *INFO* Generating Travis-CI config for testing for GHC versions: 7.8.1 7.8.2 7.8.3 7.8.4 7.10.1 7.10.2 7.10.3 8.0.1 8.0.2 8.2.1 8.2.2 8.4.1 8.4.2 8.4.3 8.4.4 8.6.1 8.6.2 8.6.3 8.6.4 8.6.5 8.8.1
   *ERROR* patch returned exit code 2
   Stdout: 
   Stderr: patch: **** Only garbage was found in the patch input.
   
   ```
3. ```
   $ ../dist-newstyle/build/x86_64-linux/ghc-8.6.4/haskell-ci-0.3.20190413/x/haskell-ci/build/haskell-ci/haskell-ci regenerate -o cabal.project.travis-patch.travis.yml 
   *INFO* Generating Travis-CI config for testing for GHC versions: 7.8.1 7.8.2 7.8.3 7.8.4 7.10.1 7.10.2 7.10.3 8.0.1 8.0.2 8.2.1 8.2.2 8.4.1 8.4.2 8.4.3 8.4.4 8.6.1 8.6.2 8.6.3 8.6.4 8.6.5 8.8.1
   *ERROR* patch returned exit code 1
   Stdout: 1 out of 1 hunk FAILED -- saving rejects to file /tmp/.travis.yml32125-0.tmp.rej
   
   Stderr: 
   
   ```

I think this is probably as good as it's going to get, so I've removed the WIP label from this PR. Final thoughts?

phadej: If travis is happy, I'll merge. Looks very good. Thanks!

src/HaskellCI.hs
```diff
@@ -217,6 +221,44 @@ genTravisFromConfigs argv config prj versions' = do
...
+    patches = cfgTravisPatches cfg
+
+    applyPatch :: FilePath -- ^ The temporary file path to patch
+               -> FilePath -- ^ The path of the .patch file
+               -> IO ()
+    applyPatch temp patch = do
+        exists <- doesFileExist patch
+        unless exists $ putStrLnErr $ "Cannot find " ++ patch
+        (ec, stdOut, stdErr) <-
+          readProcessWithExitCode "patch" [ "-i", patch
```
phadej: I'll change these to long names; long names read better.

PR #265, opened 2019-04-17
Try to sort non-cwd regenerate

phadej:
@RyanGlScott does this work for you?

### DIFF
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -33,7 +33,7 @@ import Data.Semigroup            (Semigroup (..))
 import Data.Set                  (Set)
 import Distribution.Compat.ReadP (readP_to_S)
 import Lens.Micro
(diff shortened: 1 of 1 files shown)


phadej:
relative paths are tricky, have to think.

RyanGlScott:
If I understand the code written here, then this is exactly what I want.

I'm not sure I understand your comments about relative paths. What challenges do they pose here that aren't normally present in other places where `OutputFile` is used?

phadej:
If the first .travis.yml is generated out of directory, etc sillyness.

Maybe I just merge this as is, and then refactor using `paths`

RyanGlScott:
Ah, right. It slipped my mind that `-o` has slightly different semantics in the context of `regenerate` since the file after `-o` is both an _input_ and an output.

PR #274, opened 2019-05-09
Add documentation for jobs-selection in cabal.haskell-ci

### LINKED ISSUE #273: Should `tested-with` information be specified in `cabal.haskell-ci` instead?
Currently, `haskell-ci` infers all of the information it knows about which versions of GHC to test by reading the `tested-with` stanzas of `.cabal` files. This technically works, although I find it a bit clunky for the following reasons:

1. If your repo has a `cabal.project` file with multiple `.cabal` files, `haskell-ci` requires that you list the exact same set of GHC versions in each `.cabal` file's `tested-with` stanza. This leads to a lot of ugly code duplication, such as in [here](https://github.com/ekmett/lens/blob/cd0e5f427b5d333b4ebe58e9104c3ad82746c982/lens.cabal#L15-L23), [here](https://github.com/ekmett/lens/blob/cd0e5f427b5d333b4ebe58e9104c3ad82746c982/examples/lens-examples.cabal#L19-L27), and [here](https://github.com/ekmett/lens/blob/cd0e5f427b5d333b4ebe58e9104c3ad82746c982/lens-properties/lens-properties.cabal#L16-L24).
2. Moreover, it's somewhat dubious to require that every `.cabal` file in a repo list the exact same versions of GHC in its `tested-with` stanza in the first place.

   As one example of an awkward tension that this creates, in the `primitive` repo I have to put [this list of GHC versions](https://github.com/haskell/primitive/blob/e0ff793a267f1004bb0c36590541ab58dba5168b/bench/primitive-benchmarks.cabal#L17-L25) (supporting all the way back to GHC 7.4.2) in `primitive-benchmarks.cabal`'s `tested-with` stanza. This is a complete lie, however—the `primitive` benchmarks only support GHC 7.8 onwards, a fact that is not at all obvious unless you look closely at the `build-depends`.

In light of this, perhaps we should consider allowing the range of tested GHC versions to be specified directly in `cabal.haskell-ci`? This would certainly be a change of convention for `haskell-ci`, but I think the payoff would be worth it.

RyanGlScott:
Fixes #273.

### DIFF
--- a/cabal.haskell-ci
+++ b/cabal.haskell-ci
@@ -104,6 +104,16 @@ constraint-set deepseq-1.4
 --   all: Copy every property from the cabal.project file.
 copy-fields: some
 
+-- Configure haskell-ci's behavior when presented multiple packages (e.g., a
+-- cabal.project file that specifies multiple .cabal files), each of which
+-- having its own range of GHC versions in its tested-with stanza.
+-- Possible values are:
+--   uniform: Every package must have the same range of versions in its
+--            tested-with stanza. (Default)
+--   any:     Take the union of all packages' version ranges. This implies
+--            that different packages' version ranges can be disjoint.
+jobs-selection: uniform
+
 -- Include these fields "as is" in generated cabal.project
 raw-project
   keep-going: False


phadej:
Thanks!

PR #279, opened 2019-05-30
raise upper bound on tasty

chris-martin:


phadej:
Bumped also ansi-terminal in https://github.com/haskell-CI/haskell-ci/pull/281

PR #280, opened 2019-05-30
drop lower bound on unordered-containers

chris-martin:


phadej:
Can I ask why?

I'd like to keep strict lower-bounds, as we don't test that we don't actually require something newer. Also `unordered-containers-0.2.10.0` is uploaded at Fri Jan 18 17:16:16 UTC 2019, which is already 4 months ago

chris-martin:
unordered-containers-0.2.9.0 is still the default in nixpkgs. Four months seems pretty fast; NixOS is on a six-month release cycle, and I can't imagine many other distros are much quicker.

phadej:
I think 0.2.9.0 is in nixpkgs atm.

PR #285, opened 2019-06-24
Optional packages

phadej:
ping @RyanGlScott

After second thought: supporting `optional-packages` may be handy for projects where vendored things come and go.

### DIFF
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE CPP                 #-}
 {-# LANGUAGE MultiWayIf          #-}
-{-# LANGUAGE NamedFieldPuns      #-}
 {-# LANGUAGE OverloadedLabels    #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE ScopedTypeVariables #-}
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -22,9 +22,10 @@ module HaskellCI (
 import Prelude ()
 import Prelude.Compat
 
-import Control.Monad             (forM_, liftM, unless, when)
+import Control.Monad             (forM_, unless, when)
 import Control.Monad.Catch       (MonadMask)
 import Control.Monad.IO.Class    (MonadIO (..))
+import Data.Bitraversable        (bitraverse)
 import Data.Function             (on)
 import Data.Generics.Labels ()
 import Data.List                 (groupBy, intercalate, isPrefixOf, nub, nubBy, sort, sortBy, (\\))
@@ -32,6 +33,7 @@ import Data.List.NonEmpty        (NonEmpty (..))
 import Data.Maybe                (isNothing, mapMaybe)
 import Data.Semigroup            (Semigroup (..))
 import Data.Set                  (Set)
+import Data.Void                 (Void)
 import Distribution.Compat.ReadP (readP_to_S)
 import Lens.Micro
 import System.Directory          (doesDirectoryExist, doesFileExist, setCurrentDirectory, canonicalizePath, makeRelativeToCurrentDirectory)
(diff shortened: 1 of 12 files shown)


RyanGlScott:
Thanks, @phadej! 

I'm trying to use this on `gl`, but the generated `.travis.yml` file doesn't seem to pick up the optional dependency. Do I need to enable a particular flag to use this?

phadej:
Let me check, I haven’t actually tried that :)

phadej:
@RyanGlScott when I try with `gl` repository I see

```
  - |
    echo 'packages: "."' >> cabal.project
    echo 'packages: "gl"' >> cabal.project
```

in the output

phadej:
and I only made a small changes in `gl` to get generator to work:

```diff
commit de0d63e33168f7327f97fc62abc0f951bf3419b3 (HEAD -> wip)
Author: Oleg Grenrus <oleg.grenrus@iki.fi>
Date:   Mon Jun 24 18:11:01 2019 +0300

    foo

diff --git a/gl/gl.cabal b/gl/gl.cabal
index 945b7ed..f3a3ba1 100644
--- a/gl/gl.cabal
+++ b/gl/gl.cabal
@@ -16,6 +16,7 @@ extra-source-files:
   CHANGELOG.markdown
   README.markdown
   TODO.markdown
+tested-with: GHC==8.6.5
 
 source-repository head
   type:     git
diff --git a/glgen.cabal b/glgen.cabal
index de1a425..634f18c 100644
--- a/glgen.cabal
+++ b/glgen.cabal
@@ -12,6 +12,7 @@ copyright:     Copyright © 2014-2019 Edward A. Kmett,
 category:      Graphics
 build-type:    Simple
 cabal-version: >=1.24
+tested-with: GHC==8.6.5
 extra-source-files:
   gl.xml
   man.txt
```

src/HaskellCI/Project.hs
```diff
@@ -7,31 +7,35 @@ module HaskellCI.Project (
...
-import           HaskellCI.ParsecError
+import HaskellCI.Newtypes
+import HaskellCI.Optimization
+import HaskellCI.ParsecError
 
 -- $setup
 -- >>> :seti -XOverloadedStrings
 
-data Project a = Project
+data Project b a = Project
```
RyanGlScott: Is there a reason that these type parameters are in the order `b a` rather than `a b`? It's not terribly important in the grand scheme of things, but it does look a bit strange to me.

src/HaskellCI.hs
```diff
@@ -157,13 +159,17 @@ travisFromConfigFile args opts path = do
...
     getCabalFiles
         | isNothing isCabalProject = return $ emptyProject & #prjPackages .~ [path]
         | otherwise = do
             contents <- liftIO $ BS.readFile path
-            pkgs <- either putStrLnErr return $ parseProjectFile path contents
-            over #prjPackages concat `liftM` T.mapM findProjectPackage pkgs
+            prj  <- either putStrLnErr return $ parseProjectFile path contents
+            prj' <- bitraverse findOptProjectPackage findProjectPackage prj
+            return prj'
+                { prjPackages    = concat $ prjPackages prj' ++ prjOptPackages prj'
```
RyanGlScott: It feels a bit weird to take both the `prjPackages` and `prjOptPackages` from one `Project` and combine them into a _single_ `prjPackages` list in a new `Project`, leaving the new `prjOptPackages` field empty. Is there a reason things are done this way?

phadej: yes. The only difference between`optional-packages` and `packages` is that former "globs" can match zero packages, but when the `.travis.yml` is generated we know for sure which packages should be there.

If someone removes or ads an optional package, they have to run the generation script ianyway, as it does resolving (because we want to `sdist` first).

https://cabal.readthedocs.io/en/latest/nix-local-build.html#cfg-field-optional-packages

src/HaskellCI/Project.hs
phadej: no really, but as you can see from smart coloring of GitHub diff, this way has its merits ;)

PR #291, opened 2019-08-03
Support ShellCheck 0.7.0

peti:
ShellCheck 0.7.x complains about use of "set -o pipefail" and issues an error,
saying:

    SC2039: In POSIX sh, set option pipefail is undefined.

We can (a) drop the shell command or (b) annotate it to silence ShellCheck.
I've opted for solution (a) since ShellCheck seems to have a point here.


phadej:
Thanks! I'll merge this soon.

phadej:
Incorporated your commit into https://github.com/haskell-CI/haskell-ci/pull/290

PR #303, opened 2019-08-30
Work around haskell/cabal#6214

RyanGlScott:
Due to haskell/cabal#6214, the scripts that `haskell-ci` generates are susceptible to breaking when running `v2-haddock` on `build-type: Custom` packages with certain versions of GHC (e.g., GHC 7.10.3). A fairly simple workaround is to just pass `--with-haddock <path-to-haddock>`, which this patch accomplishes.

### DIFF
--- a/src/HaskellCI/Travis.hs
+++ b/src/HaskellCI/Travis.hs
@@ -126,6 +126,9 @@ makeTravis argv Config {..} prj JobVersions {..} = do
         shForJob RangeGHCJS "HC=${HC}js"
         shForJob RangeGHCJS "WITHCOMPILER=\"--ghcjs ${WITHCOMPILER}js\""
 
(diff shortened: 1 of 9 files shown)


src/HaskellCI/Travis.hs
```diff
@@ -126,6 +126,9 @@ makeTravis argv Config {..} prj JobVersions {..} = do
         shForJob RangeGHCJS "HC=${HC}js"
         shForJob RangeGHCJS "WITHCOMPILER=\"--ghcjs ${WITHCOMPILER}js\""
 
+        -- Needed to work around haskell/cabal#6214
```
phadej: Thanks for adding reference to the issue. I usually prefer clickable links (i.e. http urls) but this is fine  for now.

PR #311, opened 2019-10-07
Build with base-compat-0.11*, generic-lens-1.2.*

RyanGlScott:


src/HaskellCI/Sh.hs
```diff
@@ -97,7 +98,7 @@ instance Applicative ShM where
     pure x = ShM $ \shs -> Right (shs, x)
     (<*>) = ap
 
-instance Monad ShM where
+instance Monad.Monad ShM where
```
phadej: hmm?

phadej: Oh, if we use `base-compat-0.11` only, I'd remove the `fail` from `Monad` definition completely.

RyanGlScott: > I'd remove the fail from `Monad` definition completely.

This would make `fail` have different semantics on old versions of `base`, as it would default to invoking `error` instead of `ShM $ \_ -> Left err`. Are you OK with this?

phadej: `haskell-ci` is `base-compat >=0.10` so `fail` exported from `base-compat-0.11` is always `Control.Monad.Fail.fail`, isn't it?

RyanGlScott: > `haskell-ci` is `base >=4.10` so `fail` exported from `base-compat-0.11` is always `Control.Monad.Fail.fail`, isn't it?

It is, but only assuming you use `Prelude.Compat` absolutely everywhere and you never make use of `MonadFailDesugaring` (which will use `Monad.fail` on pre-8.8 GHCs).

phadej: `MonadFailDesugaring` :(((

Let's remove the whole `MonadFail` instance. We have `MonadErr` for errors anyway. Then `MonadFailDesugaring` stuff will fail on 8.8.1 - and seems we don't use it.

RyanGlScott: I've removed the `{Monad,MonadFail}.fail` implementation for `ShM`.

phadej:
Thanks!

PR #314, opened 2019-10-17
GHC 8.8.1 has been released

ygale:
GHC 8.8.1 was released almost two months ago. It is supported by the HEAD version of this library, but not yet by the Hackage version. In `README.md` please remove the "pre-release" tag from GHC 8.8.1, and then upload a new version to Hackage. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ At time of writing [Travis-CI](https://travis-ci.org/) has [support for building
  - GHC 8.2.1, GHC 8.2.2
  - GHC 8.4.1, GHC 8.4.2, GHC 8.4.3, GHC 8.4.4
  - GHC 8.6.1, GHC 8.6.2, GHC 8.6.3, GHC 8.6.4, GHC 8.6.5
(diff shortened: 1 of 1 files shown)


phadej:
I'm 100% sure that https://hackage.haskell.org/package/haskell-ci-0.4 has 8.8.1, at least it is in https://hackage.haskell.org/package/haskell-ci-0.4/src/src/HaskellCI/Compiler.hs

ygale:
OK, that could be. But please bump the base constraint in the cabal file.

phadej:
The constraint is bumped in `master`. If you are forced to use GHC-8.8.1 I suggest to install from a `git clone`.

ygale:
Yes, that's what I did, and it worked great. Thanks!

phadej:
FWIW: the `ShellCheck` isn't buildable with GHC-8.8, `ShellCheck` isn't hard dependency, but it would be nice to have for GHC-8.8 builds as well - it's very handy feature, catching bugs early.

PR #321, opened 2019-11-03
First step to split out cabal-parsers

phadej:
cc @RyanGlScott


RyanGlScott:
Thanks, @phadej! I don't have any particular comments to make about the code itself, but I was able to use `cabal-parsers` in my own project without any issues, so this looks great from my perspective.

phadej:
@RyanGlScott good, I'll merge this now, but wait a bit before releasing. We need to change `SourceRepo` to the type like `cabal-install` master (allowing multiple subdirs), and I have to test-drive this package in my own project(s) before publishing.

Hopefully you are fine with using it via `source-repository-package` for about a month.

RyanGlScott:
> Hopefully you are fine with using it via `source-repository-package` for about a month.

Sounds fine to me!

PR #326, opened 2019-11-05
Fix a small bug in 277 implementation

phadej:
@RyanGlScott can you try with this branch?


RyanGlScott:
Things finally work with this branch. Thanks!

PR #327, opened 2019-11-05
Use language: shell

phadej:


phadej:
Bad idea, we need `language: c` so the $CC is passed for us. I'll add a comment about that.

PR #343, opened 2019-12-12
Add small benchmark, and NFData instances

phadej:


phadej:
Eh. `package-id bs-cmpt-bttrs-0.11.0-01c1e90f` :)

PR #358, opened 2020-01-25
Use cabal-install-3.2 with GHC 8.10 by default

### LINKED ISSUE #357: GHC 8.10.1 should use cabal-install-3.2 or later
GHC 8.10.1-rc1 ships with `Cabal-3.2.0.0`. AFAICT, this means that attempting to combine GHC 8.10.1-rc1 with `cabal-install-3.0` will cause a build plan with `Cabal-3.0.0.0` to be picked. This can actually cause problems for code that requires `Cabal-3.2` or later to build, as [this failing `singletons` Travis build](https://travis-ci.org/goldfirere/singletons/jobs/641729121#L460) demonstrates.

Unfortunately, `haskell-ci` currently picks `cabal-install-3.0` to be used with GHC 8.10.1 by default. I think the default should be `cabal-install-3.2` to avoid these sorts of nasty surprises.

RyanGlScott:
Fixes #357.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_cache:
 jobs:
   include:
     - compiler: ghc-8.10.1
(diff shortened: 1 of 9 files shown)


phadej:
Fixed in  https://github.com/haskell-CI/haskell-ci/pull/359, I noticed only the issue not the PR. Do we ask to do both?

phadej:
Fixed in  https://github.com/haskell-CI/haskell-ci/pull/359, I noticed only the issue not the PR. Do we ask to do both?

RyanGlScott:
No worries, I'm just happy that it's fixed. Thanks!

Issue #269, opened 2019-04-30
IRC notifications template is the `show` output of a string


RyanGlScott:
I noticed a [suspicious change](https://github.com/ekmett/semigroups/commit/6c3dcbed31b3ba1ef3ee66126a40758d1d8f452a#diff-354f30a63fb0907d4ad57269548329e3L21) in a recent update to `semigroups`' `.travis.yml` file:

```diff
-      - "\x0313semigroups\x03/\x0306%{branch}\x03 \x0314%{commit}\x03 %{build_url} %{message}"
-
+      - "\"\\x0313semigroups\\x03/\\x0306%{branch}\\x03 \\x0314%{commit}\\x03 %{build_url} %{message}\""
```

It appears that the previous string was fed into `show`, resulting in everything being quoted more than it should. I didn't check to see if this notification actually showed up on the #haskell-lens IRC channel, but my guess it that it wouldn't have.

RyanGlScott:
[Another part of the `.travis.yml` file](https://github.com/ekmett/semigroups/commit/6c3dcbed31b3ba1ef3ee66126a40758d1d8f452a#diff-354f30a63fb0907d4ad57269548329e3L18) actually went in the opposite direction in terms of how much quoting is present:

```diff
-      - "irc.freenode.org#haskell-lens"
+      - irc.freenode.org#haskell-lens
```

RyanGlScott:
I tried fixing this recently, but there appears to be more going on here than I originally thought. I discovered the line of code where the `notifications` template lives:

https://github.com/haskell-CI/haskell-ci/blob/2bf66a49942b1eb79c31cc71743df51d5950ebff/src/HaskellCI/Travis.hs#L382

As an first attempt to start fixing this, I made a small change to remove that first explicit double quote:

```diff
-                    [ "\"\\x0313" ++ projectName ++ "\\x03/\\x0306%{branch}\\x03 \\x0314%{commit}\\x03 %{build_url} %{message}\""
+                    [ "\\x0313" ++ projectName ++ "\\x03/\\x0306%{branch}\\x03 \\x0314%{commit}\\x03 %{build_url} %{message}\""
```

Seems simple enough, right? But here's how this change manifests in the generated `.travis.yml` file:

```diff
-      - "\"\\x0313semigroups\\x03/\\x0306%{branch}\\x03 \\x0314%{commit}\\x03 %{build_url} %{message}\""
[...]

PR #362, opened 2020-02-14
Try to fix IRC notifications template

phadej:
See https://github.com/haskell-CI/haskell-ci/issues/269#issuecomment-586318569

cc @RyanGlScott this should work now, I tried with `tree-diff``

### DIFF
--- a/src/HaskellCI/YamlSyntax.hs
+++ b/src/HaskellCI/YamlSyntax.hs
@@ -1,8 +1,8 @@
 {-# LANGUAGE DeriveFoldable      #-}
 {-# LANGUAGE DeriveFunctor       #-}
 {-# LANGUAGE DeriveTraversable   #-}
+{-# LANGUAGE GADTs               #-}
 {-# LANGUAGE ScopedTypeVariables #-}
-{-# LANGUAGE GADTs #-}
 module HaskellCI.YamlSyntax (
     Yaml (..),
     reann,
@@ -10,13 +10,12 @@ module HaskellCI.YamlSyntax (
     prettyYaml,
     ) where
 
+import HaskellCI.Prelude
 import Prelude ()
-import Prelude.Compat
 
-import Data.Bifunctor     (first)
-import Data.List.NonEmpty (NonEmpty (..))
+import Data.Bits          (shiftR, (.&.))
+import Data.Char          (isControl, isPrint, ord)
 import Data.Monoid        (Endo (..))
-import Data.String        (IsString (..))
 
 import qualified Data.Aeson              as Aeson
 import qualified Data.List.NonEmpty      as NE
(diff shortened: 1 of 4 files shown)


src/HaskellCI/YamlSyntax.hs
```diff
@@ -315,3 +318,35 @@ prettyYaml comment' = flatten . go where
 
 -- a 'Line' is comments before in and actual text after!
 data Line = Line [String] ShowS
+
+escapeYAMLString :: String -> ShowS
```
RyanGlScott: It's not clear to me how you figured out the implementation of this function, which seems crucial to the whole operation here. If there is some reference you used to figure this out, including a link to it would be helpful.

src/HaskellCI/YamlSyntax.hs
```diff
@@ -183,12 +186,12 @@ prettyYaml comment' = flatten . go where
...
     comment = concatMap lines' . comment' where
         lines' "" = [""]
-        lines' s  = lines s 
+        lines' s  = lines s
 
     go :: Yaml ann -> NonEmpty (Int, Line)
     go (YString ann s) = case literal s of
         Just ss -> pure (0, Line (comment ann) ss)
-        Nothing -> pure (0, Line (comment ann) (shows s))
+        Nothing -> pure (0, Line (comment ann) $ escapeYAMLString s)
```
RyanGlScott: There's another `escape` function (defined in `HaskellCI.Travis`) that is currently defined as:

```hs
escape :: Quotes -> String -> String
escape Double xs = show xs
```

Should this use `escapeYAMLString` as well?

src/HaskellCI/YamlSyntax.hs
phadej: I haven't found any. `\u` escaping didn't work.

src/HaskellCI/YamlSyntax.hs
phadej: No, that one would need proper `escapeSHString`...

src/HaskellCI/YamlSyntax.hs
phadej: ... Long term it would be nice to use `HsYAML` functionality, and not reinvent it.

RyanGlScott: I did find an [`escapeJSON`](https://hackage.haskell.org/package/extra-1.6.19/docs/Data-List-Extra.html#v:escapeJSON) function in the `extra` library, although that appears to be different from `escapeYAMLString` in various ways. For one thing, `escapeJSON` doesn't surround the `String` with double quotes.

If there is no reference, perhaps leave some comments highlighting what things in particular need to be escaped. The `hexChar` function in particular could benefit from this, since I have absolutely no idea what it is doing.

RyanGlScott: Thanks, the new comments help orient myself with what is going on.

PR #374, opened 2020-04-02
Require building with Cabal-3.2.* 

RyanGlScott:
As far as I can tell, there are only two changes in `Cabal-3.2.*`                              
that matter to `cabal-install-parsers`/`haskell-ci`:                                           
                                                                                               
* The type of `runParseResult` has changed:                                                    
                                                                                               
  ```diff                                                                                      
  -runParseResult :: ParseResult a -> ([PWarning], Either (Maybe Version, [PError])        a)  
  +runParseResult :: ParseResult a -> ([PWarning], Either (Maybe Version, NonEmpty PError) a)  
  ```                                                                                          

  Currently, various code paths invoke `runParseResult` and store the returned `[PError]` in the `peErrors` field of `ParseError`. To accommodate the new type signature of `runParseResult`, I have changed `ParseError` to be parameterized over the list type for `peErrors`. Most of the time, this will be instantiated to `NonEmpty`, although there is one place in `HaskellCI.ParseUtils` where this is instantiated to `[]`.
* The `FieldGrammar` class has a new method:

  ```hs
  class FieldGrammar g where
      ...
      -- | @since 3.2.0.0
      freeTextFieldDefST
          :: FieldName
          -> ALens' s ShortText -- ^ lens into the field
          -> g s ShortText
      ...
  ```

  Implementing this proves straightforward.

### DIFF
--- a/haskell-ci.cabal
+++ b/haskell-ci.cabal
@@ -137,7 +137,7 @@ library haskell-ci-internal
   build-depends:
     , base          >=4.10     && <4.15
     , bytestring    ^>=0.10.8.1
-    , Cabal         ^>=3.0
+    , Cabal         ^>= 3.2
     , containers    ^>=0.5.7.1 || ^>=0.6.0.1
     , deepseq       ^>=1.4.2.0
     , directory     ^>=1.3.0.0
--- a/cabal-install-parsers/Changelog.md
+++ b/cabal-install-parsers/Changelog.md
@@ -1,11 +1,14 @@
 ## 0.3
 
+- Require `Cabal-3.2`.
 - Rename `prjOrigFields` to `prjOtherFields` to reflect its intended purpose:
   contain all fields of a `cabal.project` file that are not already covered by
   a different field of `Project`, such as `prjPackages`, `prjOptPackages`,
   `prjSourceRepos`, etc. The semantics of `parseProject` have also been changed
   accordingly, so `Project`s produced by `parseProject` will no longer have
   `prjOtherFields` entries that overlap with other `Project` fields.
+- `ParseError` from `Cabal.Parse` now parameterizes the list type used in
+  `peErrors`. Most of the time, this will be instantiated to `NonEmpty`.
 - Add `NFData` instances
 
 ## 0.2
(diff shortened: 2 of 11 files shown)


RyanGlScott:
GitHub isn't displaying it for some reason, but the Travis build for this PR can be found here: https://travis-ci.org/github/haskell-CI/haskell-ci/builds/670135483

haskell-ci.cabal
```diff
@@ -137,7 +137,7 @@ library haskell-ci-internal
   build-depends:
     , base          >=4.10     && <4.15
     , bytestring    ^>=0.10.8.1
-    , Cabal         ^>=3.0
+    , Cabal         ^>=3.0 || ^>= 3.2
```
phadej, line 140: Just drop `3.0` support. It's easier for us.

RyanGlScott: I can do that. I have one question, however. Currently, `ParseError` is defined like this:

https://github.com/haskell-CI/haskell-ci/blob/7d6bf41f3338750b8f4b069cbd4889fe509c1e73/cabal-install-parsers/src/Cabal/Parse.hs#L50-L55

If we drop `Cabal-3.0` support, you might think that the simplest solution is to change the type of `peErrors` from `[PError]` to `NonEmpty PError`, since that would match the corresponding change in `runParseResult`'s type signature. However, that won't quite work, since there is a line of code in `haskell-ci` that specifically uses an empty list for `peErrors`:

https://github.com/haskell-CI/haskell-ci/blob/1e5d6138e51bfa90f03a33480ac8c71453153aa8/src/HaskellCI/ParsecUtils.hs#L36-L39

What should be done about this line of code?

RyanGlScott: Also, should we drop `Cabal-3.0` support for both `cabal-install-parsers` _and_ `haskell-ci`, or just `haskell-ci`?

phadej: I'd say both. As far as they are in the same repository we want be able to test that older Cabal support doesn't rot.

As for `peErrors`, I'd try to do 

```
data ParseError f = ParseError
    ...
   , peErrors = f C.PError
```

and make `renderParseError` require `Foldable f`.

If you mind trying that out, it would be great. Otherwise you can leave it as list.

RyanGlScott: I've required `Cabal-3.2` everywhere, and `ParseError` is now parameterized by `f`, the list type.

phadej:
Thanks!

phadej:
FYI, I'll update the default `cabal-install` to be `3.2` as soon as the tagged version is in HVR PPA.

Issue #383, opened 2020-04-24
{foo,bar} bad package location


philderbeast:
If I use the `{foo,bar}` syntax for package locations then haskell-ci chokes. If I instead specify the same packages individually haskell-ci is fine with that.

```
> ~/.cabal/bin/haskell-ci --output=.travis.yml --config=cabal.haskell-ci cabal.project                                 
*ERROR* Bad package location: "../{foo,bar}/"
```

This syntax is [described in the cabal docs](https://github.com/haskell/cabal/blob/3752c3e5edc3c19b7262cdb0b70f4e0539b46362/Cabal/doc/cabal-project.rst#L52-L59).

Can we please say in the readme that the brace syntax is not supported?

Are we reusing the exact same globbing code that cabal-install uses, https://github.com/haskell/cabal/issues/6732?

phadej:
Use `../{foo,bar}/*.cabal`

`haskell-ci` / `cabal-install-parsers` does cut corners, globs have to match `.cabal` files. (If I remember right they cannot be local tarballs either, as those aren't supported).

philderbeast:
I can confirm that the trailing `/*.cabal` is needed.

```
-- cabal.project
packages:
    lang-haskell/{clip,cmd,comp,detour-via-sci,detour-via-uom,earth,flare-timing,fsdb,gap,igc,kml,latlng,lookup,mask,route,scribe,siggy-chardust,span,task,tasty-compare,time,track,units,zone}/

~/.cabal/bin/haskell-ci --output=.travis.yml --config=cabal.haskell-ci cabal.project
*ERROR* Bad package location: "lang-haskell/{clip,cmd,comp,detour-via-sci,detour-via-uom,earth,flare-timing,fsdb,gap,igc,kml,latlng,lookup,mask,route,scribe,siggy-chardust,span,task,tasty-compare,time,track,units,zone}/"
```

```
-- cabal.project
packages:
    lang-haskell/{clip,cmd,comp,detour-via-sci,detour-via-uom,earth,flare-timing,fsdb,gap,igc,kml,latlng,lookup,mask,route,scribe,siggy-chardust,span,task,tasty-compare,time,track,units,zone}/*.cabal

> ~/.cabal/bin/haskell-ci --output=.travis.yml --config=cabal.haskell-ci cabal.project
*INFO* Generating Travis-CI config for testing for GHC versions: 8.2.2
```

PR #384, opened 2020-04-24
Globs need to resolve to cabal files, not folders containing them.

philderbeast:
#383


phadej:
I'll rather fix #383 properly.

PR #387, opened 2020-04-24
Typo in code comment about copy-fields.

philderbeast:


src/HaskellCI/Travis.hs
```diff
@@ -603,7 +603,7 @@ makeTravis argv Config {..} prj JobVersions {..} = do
         for_ uris $ \uri ->
             item $ C.PrettyField () "packages" $ PP.text $ uriToString id uri ""
 
-        -- copy files from original cabal.project
```
philderbeast, line 606: What in this case is the "original" cabal.project?

PR #389, opened 2020-05-06
Expand default hlint-version to include 3.0

RyanGlScott:
Now that hlint-3.0 is out, it seems like an appropriate time to widen the upper version bounds in the default `hlint-version`.

### DIFF
--- a/cabal.haskell-ci
+++ b/cabal.haskell-ci
@@ -64,7 +64,7 @@ folds: constraint-sets
 hlint: False
 hlint-yaml: .hlint.yaml
 hlint-download-binary: True
--- hlint-version: ==2.2.*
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Oh wow, `hlint-3.1` is already out. I wonder: should the version bounds encompass all of `2.2.*`, `3.0.*`, _and_ `3.1.*`? Just `2.2.*` and `3.1.*`? Something else?

phadej:
Let's wait a month. Maybe `hlint` will stabilise. The default `hlint-version` is overridable, so there is no hurry with bumping it.

phadej:
Bumped to 3.1 in https://github.com/haskell-CI/haskell-ci/pull/391 and merged. Thanks!

PR #416, opened 2020-12-12
WIP: GitHub actions

phadej:


phadej:
ping @RyanGlScott I think I have what I want to generate

.github/workflows/manual-ci.yml
```diff
@@ -0,0 +1,218 @@
...
+    branches:
+      - master
+
+jobs:
+  linux:
+    name: Haskell-CI Linux
+    strategy:
+      matrix:
+        include:
+        - image: phadej/ghc:8.10.1-bionic
```
RyanGlScott, line 28: The use of Docker here is interesting. Is there an advantage to doing this over something like [`setup-haskell`](https://github.com/actions/setup-haskell)?

src/HaskellCI/GitHub.hs
```diff
@@ -0,0 +1,122 @@
...
+import HaskellCI.GitHub.Yaml
+import HaskellCI.VersionInfo
+
+-------------------------------------------------------------------------------
+-- GitHub header
+-------------------------------------------------------------------------------
+
+githubHeader :: Bool -> [String] -> [String]
+githubHeader insertVersion argv =
+    [ "This GitHub workflow cnfig has been generated by a script via"
```
RyanGlScott, line 39:
```suggestion
    [ "This GitHub workflow config has been generated by a script via"
```

RyanGlScott (commented):
Wow! This is extremely exciting!

I've been (very slowly) working on a GHA backend in my spare time, so I'll be interested to look at this closely later and see how our approaches compare. In the meantime, I'll perform a quick review.

.github/workflows/manual-ci.yml
phadej: I'm just experimenting here. I won't use `setup-haskell` for linux builds, it doesn't really add anything (maybe marginal speed-up?)

phadej: The real case for using docker, is that with `docker` one gets much more cleaner environment, where the GitHub builder image is full of stuff and changing all the time.

RyanGlScott: Gotcha. I was only asking since I assumed that `setup-haskell` simplified the process of actually installing `ghc` and `cabal`. But if you can get GHA to do the same thing with manual invocations of `apt-get install`, then that's fine too.

.travis.yml
```diff
@@ -1,216 +0,0 @@
-# This Travis job script has been generated by a script via
```
RyanGlScott, line 1: Now that this `.travis.yml` file has been removed, would it be worth migrating the fixtures in the test suite from Travis-style YAML files to GHA-style YAML files? I did this in my version of the GHA backend, and I wonder if that would be the right thing going forward.

src/HaskellCI.hs
```diff
@@ -302,7 +305,91 @@ regenerateBash opts = do
...
+
+defaultGitHubPath :: FilePath
+defaultGitHubPath = ".github/workflows/haskell-ci.yml"
+
+doGitHub :: [String] -> FilePath -> Options -> IO ()
+doGitHub args path opts = do
+    contents <- githubFromConfigFile args opts path
+    case optOutput opts of
+        Nothing              -> do
+            createDir defaultGitHubPath
```
RyanGlScott, line 322: The code duplication among all these backends is a bit worrisome, especially since there are now subtle differences between backends (e.g., the GHA backend uses `createDir`+`writeFile`, while other backends just use `writeFile`). Is there a compelling reason not to factor out the code that is shared among all the backends?

.github/workflows/haskell-ci.yml
```diff
@@ -0,0 +1,204 @@
...
+      - name: unconstrained build
+        run: |
+          rm -f cabal.project.local
+          $CABAL v2-build $ARG_COMPILER --disable-tests --disable-benchmarks all
+      - name: prepare for constraint sets
+        run: |
+          rm -f cabal.project.local
+      - name: constraint set deepseq-1.4
+        run: |
+          if [ $((HCNUMVER < 80400)) -ne 0 ] ; then $CABAL v2-build $ARG_COMPILER --disable-tests --disable-benchmarks --constraint='deepseq ==1.4.*' --constraint='binary installed' all ; fi
```
RyanGlScott, line 203: This isn't something that needs to be done right now, but I wonder if it would be cleaner to replace these `bash` `if`-expressions with GHA's [`jobs.<job_id>.if`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif).

.travis.yml
phadej: I'm not sure what to do. Technically I don't want to drop Travis support. But if we (`haskell-ci` org) run out of credits, so we cannot self test, it's a bit of a problem.

But in general, yes. We should test all backends across all fixtures. The test-suite might need some reorganization to not make it all a complete mess :)

src/HaskellCI.hs
phadej: I don't find it yet problematic. The code duplication in actual generation is something which we need to dedup, but there as well we should do it only when we feel the pain, not preemptively. My previous attempts to introduce extra backends failed because I tried to generalize to early.

.github/workflows/haskell-ci.yml
phadej: Technically yes. But right now the `sh` code is almost the same across at least `bash` and `github` backends.

.github/workflows/manual-ci.yml
phadej: It would "simplify" for macOS where one don't have anything like Docker so (unfortunately) have to run on GitHub provided image, then `setup-haskell` would be able to use pre-installed GHCs.

Note, that installing with `apt` takes about 30 seconds, i.e. barely anything in larger scheme of things. On macOS `ghcup` uses bindists which `configure` and `make install` take noticeable time. So there it matters.

.github/workflows/haskell-ci.yml
phadej: Also not *all* `if` can be disabled. (E.g. when we have different packages per different GHCs enabled).

.travis.yml
RyanGlScott: > But in general, yes. We should test all backends across all fixtures. The test-suite might need some reorganization to not make it all a complete mess :)

In my `wip/github-actions` branch, I include both `.travis.yml` and `github-actions.yml` files for each test fixture (see [this commit](https://github.com/RyanGlScott/haskell-ci/commit/00d793b965298c57774174c5ccbcc1c805a6ae91)). Is this what you had in mind?

phadej: Yes.

And you can see yourself `fixtures` directory is complete mess. Doesn't invite to add any more tests :)

RyanGlScott: OK. I was going to ask if you'd like me to create a PR to add these `github-actions.yml` files to the test suite. Perhaps you'd prefer to wait until the test suite is in a cleaner state first?

phadej: Yeah. I have a thing in mind which would allow making fixtures situation a bit nicer. Will do that soon before I forget that.

.github/workflows/manual-ci.yml
jared-w: haskell/actions/setup (formerly setup-haskell) takes care of a few things:

1. Cross platform support with the same syntax. This is actually what I consider to be the most important aspect of it. Specifying "I want this version of GHC and this version of cabal-install on my path" with the same syntax regardless of the underlying OS is very powerful.
2. Caching, speed, performance. This is all really nice stuff, but not necessarily essential; if you already have the right version of GHC+cabal-install setup in your PATH, the action will no-op. This allows you to do things like use `haskell/actions/setup` multiple times in a single job and pay the most minimal cost possible. It also makes using it a zero cost abstraction for self hosted runners. To put it another way, I go to pretty far lengths to make sure there's "no reason" to avoid using it.
3. Integration with github actions. This is another "real"/main benefit of using it. As it always runs natively, you don't have to worry about docker/filesystem abstractions leaking. It also takes care to not do anything "unexpected", so it will always compose well with other github actions; it is much easier to chain `haskell/actions/setup` with something like `actions/create-release`, for example.
   It also sets convenience outputs so that you can reference, eg, the cabal store (or stack root) in a platform-agnostic way for setting up caching easier or whatever you want.
   Going further, `haskell/actions/setup` could integrate things like options to automatically cache dist-newstyle using something like `cabal-cache`, or setting up "matchers" so that GHC/cabal/etc's errors or warnings had tighter integration with github action's UI.

As for what that looks like, implementation wise: for linux, it checks to see if GHC and/or cabal-install was already installed. If not, it'll try hvr's PPA and use that. If that doesn't work, it'll download `ghcup` and use that to install things. If stack is enabled, it'll do the same thing for stack (check if installed and download from github releases if needed). Lastly, it sets the PATH, and outputs a few convenience things like the location of cabal's store or the stack root.

So, yes, one can easily do this on linux themselves by just blindly installing GHC using hvr's ppa and the setting the path themselves. It will take slightly more time for pre-installed GHC's than otherwise because apt-get will be invoked rather than just setting the path. On the other hand, that time might be washed out by github actions actually cloning this repo and then running the right node program. So... meh; it's kind of a wash if you don't care about any of the other things `haskell/actions/setup` offers.

In the end, whether all of that *matters* for `haskell-ci` is an entirely different question. If `haskell-ci` wants to enable cross platform testing conveniently, it will eventually end up re-implementing the majority of `haskell/actions/setup`'s logic, so it will make sense to integrate other tooling where possible (from a sheer effort perspective), but that might not take the form of actually using `haskell/actions/setup`, particularly if the interest is keeping the exact same logic across different CI platforms + locally.

phadej: > Lastly, it sets the PATH

By the way, that is a thing which `haskell-ci` **does not** do. There is no `cabal` or `ghc` in `PATH`. That breaks some "fihsy" stuff where authors depends on compiler to be `ghc` and in PATH. (They could use `ghc-paths` which at least explicitly declares that kind of nastyness).

arianvp:
Thank you SO MUCH for this!

.github/workflows/manual-ci.yml
jared-w: > By the way, that is a thing which `haskell-ci` **does no**t do.

Yeah, it's interesting. That approach works well when you have the entire haskell testing setup self contained, but the way github actions tend to be structured is by starting with a clean environment, mutating the environment to setup, and then running the tests. So the action needs to add relevant programs to the PATH to integrate well. I do also export paths for the directory + binary for each of ghc, stack, and cabal-install as outputs, but I have never actually seen them used.

Interestingly, I can't actually wipe the path inside an action. The design of `echo FOO=bar >> $GITHUB_PATH` prevents this (intentionally). So I "have" to do that design. `haskell-ci` could, of course, generate a script where the first part of every step resets/wipes the PATH of all haskell related paths.

andreasabel:
@phadej Phantastic job! Tried it on my first project and it worked out of the box!

PR #418, opened 2020-12-14
Add --github-patches counterpart to --travis-patches

RyanGlScott:
I found myself needing this flag in certain configurations of mine.

### DIFF
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -164,44 +164,6 @@ genTravisFromConfigs argv config prj vs = do
                 , "# EOF"
                 ]
 
--- | Adjust the generated Travis YAML output with patch files, if specified.
--- We do this in a temporary file in case the user did not pass --output (as
--- it would be awkward to patch the generated output otherwise).
-patchTravis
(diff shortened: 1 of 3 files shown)


src/HaskellCI.hs
```diff
@@ -164,25 +164,32 @@ genTravisFromConfigs argv config prj vs = do
                 , "# EOF"
                 ]
 
--- | Adjust the generated Travis YAML output with patch files, if specified.
--- We do this in a temporary file in case the user did not pass --output (as
--- it would be awkward to patch the generated output otherwise).
 patchTravis
```
phadej, line 167: These could be moved out of

```

-------------------------------------------------------------------------------
-- Travis
-------------------------------------------------------------------------------
```

into own 

```

-------------------------------------------------------------------------------
-- Patches
-------------------------------------------------------------------------------
```

phadej (changes requested):

src/HaskellCI.hs
RyanGlScott: Good idea! I've applied your suggestion.

phadej:
Thanks!

PR #425, opened 2020-12-16
GitHub Actions: support --allow-failures through continue-on-error

### LINKED ISSUE #424: GitHub Actions backend: support --allow-failures
Unfortunately, GitHub Actions doesn't support a Travis-esque/GitLab CI-esque notion of `allow-failures` out of the box (see actions/toolkit#399), but it does have [`jobs.<job_id>.continue-on-error`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error), which is probably the closest thing we'll get.

RyanGlScott:
The UX for `continue-on-error` is not the greatest, but it's the closest thing we have to `allow-failures` in GitHub Actions.

Fixes #424.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -24,8 +24,11 @@ jobs:
     runs-on: ubuntu-18.04
     container:
       image: buildpack-deps:bionic
+    continue-on-error: ${{ matrix.continue-on-error }}
     strategy:
       matrix:
+        continue-on-error:
+          - false
         include:
           - ghc: 8.10.1
           - ghc: 8.8.3
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -20,7 +20,7 @@ on:
       - master
 jobs:
   linux:
-    name: Haskell-CI Linux
+    name: Haskell-CI Linux - GHC ${{ matrix.ghc }}
     runs-on: ubuntu-18.04
     container:
       image: buildpack-deps:bionic
--- a/src/HaskellCI/GitHub.hs
+++ b/src/HaskellCI/GitHub.hs
@@ -353,12 +353,13 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
             , ghjSteps     = steps
             , ghjContainer = Just "buildpack-deps:bionic" -- use cfgUbuntu?
             , ghjMatrix    =
-                [ Map.fromList
-                    [ ("ghc", v')
-                    -- , ("isGhcjs", "0")
-                    ]
-                | GHC v <- reverse $ toList versions
-                , let v' = prettyShow v
+                [ GitHubMatrixEntry
+                    { ghmeGhcVersion = v
+                    , ghmeContinueOnError =
+                           previewGHC cfgHeadHackage compiler
+                        || maybeGHC False (`C.withinRange` cfgAllowFailures) compiler
+                    }
+                | compiler@(GHC v) <- reverse $ toList versions
                 ]
             }
         }
--- a/src/HaskellCI/GitHub.hs
+++ b/src/HaskellCI/GitHub.hs
@@ -348,7 +348,7 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
             { ghBranches = cfgOnlyBranches
             }
         , ghJobs = Map.singleton "linux" GitHubJob
-            { ghjName      = "Haskell-CI Linux"
+            { ghjName      = "Haskell-CI Linux - GHC ${{ matrix.ghc }}"
             , ghjRunsOn    = "ubuntu-18.04" -- TODO: use cfgUbuntu
             , ghjSteps     = steps
             , ghjContainer = Just "buildpack-deps:bionic" -- use cfgUbuntu?
--- a/src/HaskellCI/GitHub/Yaml.hs
+++ b/src/HaskellCI/GitHub/Yaml.hs
@@ -30,8 +30,14 @@ data GitHubJob = GitHubJob
     { ghjName      :: String
     , ghjRunsOn    :: String
     , ghjContainer :: Maybe String
-    , ghjMatrix    :: [M.Map String String]
+    , ghjMatrix    :: [GitHubMatrixEntry]
     , ghjSteps     :: [GitHubStep]
+    }
+  deriving (Show)
+
+data GitHubMatrixEntry = GitHubMatrixEntry
+    { ghmeGhcVersion      :: Version
+    , ghmeContinueOnError :: Bool
     }
   deriving (Show)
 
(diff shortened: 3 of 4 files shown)


src/HaskellCI/GitHub.hs
```diff
@@ -349,7 +349,7 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
             { ghBranches = cfgOnlyBranches
             }
         , ghJobs = Map.singleton "linux" GitHubJob
-            { ghjName      = "Haskell-CI Linux"
+            { ghjName      = "Haskell-CI Linux - GHC ${{ matrix.ghc }}"
```
phadej, line 352: Have you found a reference of that templating syntax anywhere, or are just copying the examples (that what I have done so far).

I wonder what happens when we want to add GHCJS support.

src/HaskellCI/GitHub/Yaml.hs
```diff
@@ -30,7 +30,7 @@ data GitHubJob = GitHubJob
     { ghjName      :: String
     , ghjRunsOn    :: String
     , ghjContainer :: Maybe String
-    , ghjMatrix    :: [M.Map String String]
+    , ghjMatrix    :: [M.Map String (Yaml [String])]
```
phadej, line 33: These types shouldn't reference `Yaml`.

src/HaskellCI/GitHub.hs
RyanGlScott: > Have you found a reference of that templating syntax anywhere

Is [this](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions) what you're looking for?

> I wonder what happens when we want to add GHCJS support.

In that case, I imagine that we'd need to switch from using `ghc: X.Y.Z` to something like this:

```yaml
- compiler-version: 8.10.2
  compiler: "GHC"
- compiler-version: 8.10.2
  compiler: "GHCJS"
```

And then change `name` to be something like `"Haskell-CI Linux - ${{matrix.compiler}} ${{matrix.compiler-version}}"`. Disclaimer: I haven't thought through the ramifications of this design thoroughly.

src/HaskellCI/GitHub/Yaml.hs
RyanGlScott: What is the alternative? `ghc` needs to map to a `String`, while `continue-on-error` needs to map to a `Bool`.

phadej: new datatype!

RyanGlScott: I have added a new `GitHubMatrixValue` with two constructors, one for `String` and one for `Bool`. Is this what you had in mind?

src/HaskellCI/GitHub/Yaml.hs
```diff
@@ -30,11 +30,17 @@ data GitHubJob = GitHubJob
     { ghjName      :: String
     , ghjRunsOn    :: String
     , ghjContainer :: Maybe String
-    , ghjMatrix    :: [M.Map String String]
+    , ghjMatrix    :: [M.Map String GitHubMatrixValue]
```
phadej, line 33: No, I think the whole `M.Map String` is a bit to dynamic to begin with. (I was lazy, sorry).

```haskell
ghcMatirx :: [GitHubMatrixEntry]

data GitHubMatrixEntry = GitHubMatrixEntry
    { ghmeGhcVersion :: String -- or Version?
    , ghcmeContinueOnError :: Bool
    }
```

RyanGlScott: Alright, I've incorporated your suggestion.

.github/workflows/haskell-ci.yml
```diff
@@ -20,12 +20,15 @@ on:
...
-    name: Haskell-CI Linux
+    name: Haskell-CI Linux - GHC ${{ matrix.ghc }}
     runs-on: ubuntu-18.04
     container:
       image: buildpack-deps:bionic
+    continue-on-error: ${{ matrix.continue-on-error }}
     strategy:
       matrix:
+        continue-on-error:
+          - false
```
phadej, line 31: the code implies `false` case shouldn't be generated?

RyanGlScott: Right, the presentation that GitHub Actions uses for this is a bit confusing. Supposing we have this:

```yaml
matrix:
  continue-on-error:
    - false
  include:
    - ghc: 8.10.1
    - ghc: 8.8.3
      continue-on-error: true
```

Then the default `continue-on-error` setting will be `false`, and that setting will be inherited for each entry in `include` unless they explicitly override `continue-on-error` (e.g., the `ghc: 8.8.3` entry). If you wanted to be more explicit, you could also do this:

```yaml
matrix:
  include:
    - ghc: 8.10.1
      continue-on-error: false
    - ghc: 8.8.3
      continue-on-error: true
```

I wanted to avoid repeating the line `continue-on-error: false` over and over again, so I didn't opt for this approach. But if you'd prefer that presentation, I could just as well switch to that.

phadej: Don't you need 

```yaml
matrix:
  continue-on-error: ${{ matrix.continue-on-error }}
  include:
    - ghc: 8.10.1
      continue-on-error: false
    - ghc: 8.8.3
      continue-on-error: true
```

I was quite sure that includes aren't automatically incorporated. Or is `continue-on-error` one of rare keys which are?

RyanGlScott: If you specify `continue-on-error` for each `include` entry, then you don't need a `continue-on-error` entry under `matrix` itself. See [this commit](https://github.com/RyanGlScott/gha-sandbox/commit/73448f81a04679586f3e661efeded375c355e835), as well as its [associated workflow](https://github.com/RyanGlScott/gha-sandbox/actions/runs/426417388).

Note that when you reference `${{ matrix.continue-on-error }}`, you're referencing the value of `continue-on-error` for the particular `include` entry that some job in a matrix is using. As a result, it doesn't really make sense to reference `${{ matrix.continue-on-error }}` inside of the matrix itself.

phadej: Oh, yes sorry. So there is `${{ matrix.continue-on-error }}`` but it's one level up.

```yaml
continue-on-error: ${{ matrix.continue-on-error }}
matrix:
  include:
    - ghc: 8.10.1
      continue-on-error: false
    - ghc: 8.8.3
      continue-on-error: true
```

OK. I think I understand how things compose here.

And as `continue-on-error` is something I don't expect to be used much, your way of not having it per each `include` is better.

RyanGlScott:
Oh dear, I appear to have subtly broken something with this patch. After these changes, GitHub Actions now only runs one job at most per workflow! See [here](https://github.com/haskell-CI/haskell-ci/actions/runs/426436640) for an example.

RyanGlScott:
Specifically, this appears to happen when none of the entries under `include` specify a value for `continue-on-error`. Compare the workflows [before](https://github.com/RyanGlScott/gha-sandbox/actions/runs/426641435) and [after](https://github.com/RyanGlScott/gha-sandbox/actions/runs/426643038) this [commit](https://github.com/RyanGlScott/gha-sandbox/commit/fc405c170fd6fdf79e771f4f264b6a5f566c4572):

```diff
diff --git a/.github/workflows/haskell-ci.yml b/.github/workflows/haskell-ci.yml
index 9f26f87..2d5bf8a 100644
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -29,8 +29,7 @@ jobs:
           - false
         include:
           - ghc: 8.10.2
-          - continue-on-error: true
-            ghc: 8.8.4
+          - ghc: 8.8.4
       fail-fast: false
     steps:
       - name: apt
```

It's unclear to me why this (seemingly buggy) behavior occurs, but I suppose this means that we'll have to specify `continue-on-error` in each `include` entry after all. Sigh. I'll prepare a PR for that.

PR #427, opened 2020-12-17
GitHub Actions: Add support for HLint

RyanGlScott:
I cargo-culted this code from `HaskellCI.Travis`, as is the case for most code in `HaskellCI.GitHub`, and tested it [here](https://github.com/RyanGlScott/gha-sandbox/runs/1570951250). Let me know if you would prefer this code be factored out more.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -103,7 +103,7 @@ jobs:
         uses: actions/cache@v2
         with:
           key: ${{ runner.os }}-${{ matrix.ghc }}-tools
-          path: ~/.cabal/store-tools
+          path: ~/.haskell-ci-tools
       - name: install cabal-plan
         run: |
           mkdir -p $HOME/.cabal/bin
@@ -114,7 +114,7 @@ jobs:
           chmod a+x $HOME/.cabal/bin/cabal-plan
       - name: install doctest
         run: |
-          $CABAL --store-dir=$HOME/cabal/store-tools v2-install $ARG_COMPILER --ignore-project -j2 doctest --constraint='doctest ^>=0.17'
+          $CABAL --store-dir=$HOME/.haskell-ci-tools/store v2-install $ARG_COMPILER --ignore-project -j2 doctest --constraint='doctest ^>=0.17'
           doctest --version
       - name: checkout
         uses: actions/checkout@v2
--- a/src/HaskellCI/GitHub.hs
+++ b/src/HaskellCI/GitHub.hs
@@ -159,7 +159,7 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
 
         when (doctestEnabled || cfgHLintEnabled cfgHLint) $ githubUses "cache (tools)" "actions/cache@v2"
             [ ("key", "${{ runner.os }}-${{ matrix.ghc }}-tools")
-            , ("path", "~/.cabal/store-tools")
+            , ("path", "~/.haskell-ci-tools")
             ]
 
         githubRun "install cabal-plan" $ do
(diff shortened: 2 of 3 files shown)


.github/workflows/haskell-ci.yml
```diff
@@ -102,8 +102,11 @@ jobs:
       - name: cache (tools)
         uses: actions/cache@v2
         with:
-          key: ${{ runner.os }}-${{ matrix.ghc }}-tools
-          path: ~/.cabal/store-tools
+          key: |
+            ${{ runner.os }}-${{ matrix.ghc }}-tools
+          path: |
+            ~/.cabal/store-tools
+            ~/.hlint
```
phadej, line 109: I'd download to `~/.cabal/store-tools`. That was an idea. The name is bad though. Rather it could be `~/.haskell-ci-tools`

The  `doctest` installation should probably be changed to use `~/.haskell-ci-tools/store` as store, and hlint could be in `~/.haskell-ci-tools/hlint`.

src/HaskellCI/GitHub/Yaml.hs
```diff
@@ -139,6 +139,12 @@ mapToYaml m = ykeyValuesFilt []
     | (k, v) <- M.toList m
     ]
 
+multiMapToYaml :: M.Map String [String] -> Yaml [String]
+multiMapToYaml m = ykeyValuesFilt []
```
phadej, line 143: This is misleading. From the name one would expect this to generate a list, and not use `unlines`. Better to not use multiple paths at all

phadej (changes requested):

.github/workflows/haskell-ci.yml
RyanGlScott: I'd be fine with that.

Tangentially related question: when we `cabal v2-install` `doctest` and `hlint`, why do we specify a separate `--store-dir` in the first place? In particular, the Travis and Bash backends don't do this, so I'm wondering why the GitHub Actions backend has this additional complexity.

src/HaskellCI/GitHub/Yaml.hs
RyanGlScott: Fine with me. I only did this so that I could pass multiple things to `path:` in the `cache` action, but if we're going to lump everything under `~/.haskell-ci-tools`, then I don't need to do this in the first place.

.github/workflows/haskell-ci.yml
phadej: bash script has a luxury of using single store for everything. Doesn't need to worry about uploads/downloads.

If you look closely, the GitHub actions uses two caches,
1. This one for tools which has strict key.
2. Another for actual project

Important bit to know is that GitHub actions **doesn't** re-upload  cache if it's downloaded with exact hit of the key. 
(n fact, we should add something identifying hlint/doctest versions there for tools-cache).

The actual dependencies store cache is separate, and is (unfortunately) completely re-uploaded for every job. We use `restore-key` to get *some* cache, there is prefix-matching logic. There are options, currently we use `${{ github.sha }}` which is globally available. Another alternative is to generate `cabal.project.freeze` and use its hash as cache-key. The latter would be better (then in PRs caches won't be reuploaded, saves some time and bandwidth). But we cannot caclculate cabal.project.freeze before checkout, and tools are installed before that. So that is why two separate caches.

phadej: Also tools cache is indeed very static, so it makes sense to have it separate. (If not `doctest` dependendcy on GHC, it could be be shared across all jobs!)

RyanGlScott: Good to know, thanks.

I've pushed an updated version that uses `~/.haskell-ci-tools`. Can you double-check to make sure I didn't do something silly in the process of changing the store location?

src/HaskellCI/GitHub.hs
```diff
@@ -304,7 +324,23 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
...
+        -- hlint
+        when (cfgHLintEnabled cfgHLint) $ githubRun "hlint" $ do
+            let "" <+> ys = ys
+                xs <+> "" = xs
+                xs <+> ys = xs ++ " " ++ ys
+
+                prependSpace "" = ""
+                prependSpace xs = " " ++ xs
+
+            let hlintOptions = prependSpace $ maybe "" ("-h ${TOP}/" ++) (cfgHLintYaml cfgHLint) <+> unwords (cfgHLintOptions cfgHLint)
```
phadej, line 336: there is no `TOP`. It should be `$GITHUB_WORKSPACE/source` (that's where we checkout the packages).

phadej (commented):
Looks fine.

In fact, we have `.hlint.yaml` for `haskell-ci`. I don't know why we don't self-test it. That would be nice. Except that `hlint` very unhappy with `haskell-ci` current state.

Let me make HLint happy, so I can then continue tweaking this PR.

Thanks for starting this.

phadej:
Merged in https://github.com/haskell-CI/haskell-ci/pull/429, thanks!

PR #432, opened 2020-12-18
WIP: GitHub Actions: Add support for --irc-channels

### LINKED ISSUE #423: GitHub Actions backend: support --irc-channels
There are a variety of existing GitHub Actions ([example](https://github.com/marketplace/actions/notify-irc)) that allow notifying an IRC channel based on various events. We should explore this and find a combination of events that gives us roughly equivalent functionality to what `--irc-channels` does for the Travis backend.

RyanGlScott:
This leverages https://github.com/marketplace/actions/irc-message-action to send IRC notifications. Since you have to manually specify which events trigger notifications, I'm restricting this to simple "The build passed" and "The build failed" messages, but it would be possible to extend this to other events later.

Fixes #423.


phadej:
As now it's a bit simpler to add new test-cases, could you add one which has `IrcChannels` config?

RyanGlScott:
Sure! See `fixtures/irc-channels.*`.

src/HaskellCI/GitHub.hs
```diff
@@ -480,6 +488,47 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
...
+    ircJob :: [String] -> ListBuilder (String, GitHubJob) ()
+    ircJob serverChannelNames = item ("irc", GitHubJob
+        { ghjName            = "Haskell-CI (IRC notification)"
+        , ghjRunsOn          = "ubuntu-18.04"
+        , ghjNeeds           = [mainJobName]
+        , ghjIf              = Nothing
+        , ghjContainer       = Nothing
+        , ghjContinueOnError = Nothing
+        , ghjMatrix          = []
+        , ghjSteps           = concatMap (\serverChannelName -> [ ircStep serverChannelName True
```
phadej, line 500: For the future, I usually use list comprehensions instead of `concatMap`, specially here, where you still need to give element a name,

src/HaskellCI/GitHub.hs
```diff
@@ -480,6 +488,47 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
     withHaddock :: String
     withHaddock = "--with-haddock $HADDOCK"
 
+    ircJob :: [String] -> ListBuilder (String, GitHubJob) ()
+    ircJob serverChannelNames = item ("irc", GitHubJob
+        { ghjName            = "Haskell-CI (IRC notification)"
+        , ghjRunsOn          = "ubuntu-18.04"
+        , ghjNeeds           = [mainJobName]
+        , ghjIf              = Nothing
```
phadej, line 496: This is where I can include `Just "${{ github.repository == 'ekmett/lens' }}"`?

src/HaskellCI/GitHub.hs
RyanGlScott: I've rewritten this to use a list comprehension.

src/HaskellCI/GitHub.hs
RyanGlScott: Yes. That being said, I should probably test this claim once it's implemented.

phadej:
Merged in #434

PR #434, opened 2020-12-19
GitHub Actions: Add support for --irc-channels

### LINKED ISSUE #423: GitHub Actions backend: support --irc-channels

phadej:
This leverages https://github.com/marketplace/actions/irc-message-action
to send IRC notifications. Since you have to manually specify which events
trigger notifications, I'm restricting this to simple "The build passed"
and "The build failed" messages, but it would be possible to extend this to
other events later.

Fixes #423.

---

This is #432, rebased (conflict resolved) and squashed.


src/HaskellCI/GitHub.hs
```diff
@@ -394,23 +394,31 @@ makeGitHub _argv config@Config {..} prj jobs@JobVersions {..} = do
...
-                | compiler@(GHC v) <- reverse $ toList versions
-                ]
-            }
+        , ghJobs = Map.fromList $ buildList $ do
+            item (mainJobName, GitHubJob
+                { ghjName            = "Haskell-CI Linux - GHC ${{ matrix.ghc }}"
+                , ghjRunsOn          = "ubuntu-18.04" -- TODO: use cfgUbuntu
+                , ghjNeeds           = []
+                , ghjSteps           = steps
+                , ghjIf              = Nothing -- TODO: Use cfgIrcIfInOriginRepo
```
RyanGlScott, line 403: Should we file an issue about this remaining task?

RyanGlScott (commented):
Thanks!

src/HaskellCI/GitHub.hs
RyanGlScott: Ah, I think this was done in #436.

phadej: Yeah, writing a patch felt easier than opening an issue.

PR #441, opened 2020-12-21
GHA: Use always() to ensure IRC job is always run

### LINKED ISSUE #437: GHA: irc notifications are not issued if build job fails
Due dependency, the notification job is simply not run...

https://github.com/haskell-CI/haskell-ci/runs/1585115378?check_suite_focus=true

RyanGlScott:
Fixes #437.


phadej:
Thanks!

PR #443, opened 2020-12-21
Fix #440: Accept more urls as github origin

### LINKED ISSUE #440: GHA: --irc-if-in-origin-repo's parsing of GitHub URLs is fragile
If I check out the `haskell-ci` repo at commit 8629a90db15ea453d65c6d20e94de75672fb41b4:

```
$ git clone https://github.com/haskell-CI/haskell-ci
$ git reset --hard 8629a90db15ea453d65c6d20e94de75672fb41b4
```

And then regenerate `.github/workflows/haskell-ci.yml`:

```
$ cabal run haskell-ci -- regenerate
```

Then the contents of `.github/workflows/haskell-ci.yml` will change!

```
$ git diff
diff --git a/.github/workflows/haskell-ci.yml b/.github/workflows/haskell-ci.yml
index 0979ac5..7629b83 100644
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -24,7 +24,6 @@ jobs:
     runs-on: ubuntu-18.04
     needs:
       - linux
-    if: ${{ github.repository == 'haskell-CI/haskell-ci' }}
     strategy:
       fail-fast: false
     steps:
```

It seems strange that this line in particular would be deleted, especially since `haskell-ci` enables the `irc-if-in-origin-repo` option. But things get even weirder from here. Let's try using a different URL and regenerating:

```
$ git remote set-url origin git@github.com:haskell-CI/haskell-ci.git
$ cabal run haskell-ci -- regenerate
```

All of a sudden, the deleted line is restored! That is, `git diff` will report no changes.

This appears to be because the logic that `haskell-ci` uses to parse `git` URLs assumes very particular conventions:

https://github.com/haskell-CI/haskell-ci/blob/8629a90db15ea453d65c6d20e94de75672fb41b4/src/HaskellCI/GitHub.hs#L564-L569

phadej:
cc @RyanGlScott

### DIFF
--- a/src/HaskellCI/GitHub.hs
+++ b/src/HaskellCI/GitHub.hs
@@ -8,6 +8,8 @@ module HaskellCI.GitHub (
 
 import HaskellCI.Prelude
 
+import Control.Applicative (optional)
+
 import qualified Crypto.Hash.SHA256              as SHA256
 import qualified Data.Attoparsec.Text            as Atto
 import qualified Data.Binary                     as Binary
@@ -41,6 +43,9 @@ import HaskellCI.Sh
 import HaskellCI.ShVersionRange
 import HaskellCI.Tools
 import HaskellCI.VersionInfo
+
+-- $setup
+-- >>> :set -XOverloadedStrings
 
 -------------------------------------------------------------------------------
 -- GitHub header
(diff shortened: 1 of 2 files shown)


src/HaskellCI/GitHub.hs
```diff
@@ -556,16 +561,37 @@ cat path contents = sh $ concat
     , "EOF"
     ]
 
+-- |
```
RyanGlScott, line 564: It might be worth mentioning here that this will likely not parse all possible Git URLs, but it should cover most common use cases.

src/HaskellCI/GitHub.hs
```diff
@@ -556,16 +561,37 @@ cat path contents = sh $ concat
...
     sshP = do
         _ <- Atto.string "git@github.com:"
         repo <- Atto.takeWhile (/= '.')
-        _ <- Atto.string ".git"
+        _ <- optional (Atto.string ".git")
+        return repo
+
+    httpsP :: Atto.Parser Text
+    httpsP = do
+        _ <- Atto.string "https://github.com/"
```
RyanGlScott, line 594: Speaking of common use cases, I think it might be worthwhile to accept `git://` as a possible prefix, if only because the official Git documentation [uses that as an example](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes).

src/HaskellCI/GitHub.hs
phadej: Git**Hub** urls.. :)

phadej: I added a comment

RyanGlScott (approved):

src/HaskellCI/GitHub.hs
RyanGlScott: Yes, good point.

Issue #385, opened 2020-04-24
Add --tested-with-set, difference or intersection


philderbeast:
Could we please add a new option that defines this set comparison, keeping the default as it is?

https://github.com/haskell-CI/haskell-ci/blob/3a6929e6b3101b6cbad56f9097e9c9bff106bbe9/src/HaskellCI/TestedWith.hs#L68

```
> haskell-ci --help
...
--tested-with-set SET  Set comparision (difference, intersection) (Default: difference)
                       of tested-with compiler versions among packages.

                       With difference, reject an non-empty difference.

                       With intersection, accept a non-empty intersection.
```

```
# alternative wording
--tested-with-set SET  Set comparision (difference, intersection) (Default: difference)

                       With difference, tested-with has to be the same for all packages.

                       With intersection, use the subset of tested-with compiler versions
                       common to all packages.
```

philderbeast:
A relevant test:

https://github.com/haskell-CI/haskell-ci/blob/0305f597a3227351671d9508f31ef19612307268/fixtures/cabal.project.fail-versions.stderr#L1-L3

phadej:
are you sure you won't be happy with `jobs-selection: any`?

philderbeast:
It is good to know about `jobs-selection: any`, thanks. I tried it.

```cabal
tested-with: GHC ==8.2.2
packages:
    lang-haskell/hcoord/{hcoord,hcoord-utm}/*.cabal
    lang-haskell/{clip,cmd,comp,detour-via-sci,detour-via-uom,earth,flare-timing,fsdb,gap,igc,kml,latlng,lookup,mask,route,scribe,siggy-chardust,span,task,tasty-compare,time,track,units,zone}/*.cabal

~/.cabal/bin/haskell-ci --output=.travis.yml --config=cabal.haskell-ci cabal.project --jobs-selection any
*INFO* Generating Travis-CI config for testing for GHC versions: 8.2.2 8.4.4 8.6.5 8.8.3
*INFO* hcoord-utm : 8.2.2 8.4.4 8.6.5 8.8.3
*INFO* hcoord : 8.2.2 8.4.4 8.6.5 8.8.3
*INFO* flight-igc : 8.2.2
*INFO* flight-cmd : 8.2.2
*INFO* detour-via-sci : 8.2.2
*INFO* flight-track : 8.2.2
*INFO* flight-route : 8.2.2
*INFO* flight-clip : 8.2.2
[...]

PR #444, opened 2020-12-22
UX: Improve discoverability of --jobs-selection any

andreasabel:
This PR adds hint
```
  (Consider option --jobs-selection any)
```
to
```
  *ERROR* FOO is missing tested-with annotation for: ghc-x.y.z
```
To deal with linebreaks in errors in the golden testsuite, I had to make the `cmp` function more robust.  Now it first normalizes the comparatees wrt. line breaks via the normalization function
```
  unlines . lines
```
See e.g. https://github.com/haskell-CI/haskell-ci/issues/385#issuecomment-619156827 why this UX improvement is helpful.

### DIFF
--- a/fixtures/fail-versions.github
+++ b/fixtures/fail-versions.github
@@ -1,3 +1,4 @@
 # FAILURE
+# *INFO* Consider option --jobs-selection any
 # *ERROR* servant-client-core is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
 # *ERROR* servant-foreign is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
--- a/fixtures/fail-versions.travis
+++ b/fixtures/fail-versions.travis
@@ -1,3 +1,4 @@
 # FAILURE
+# *INFO* Consider option --jobs-selection any
 # *ERROR* servant-client-core is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
 # *ERROR* servant-foreign is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
--- a/fixtures/fail-versions.bash
+++ b/fixtures/fail-versions.bash
@@ -1,3 +1,4 @@
 # FAILURE
+# *INFO* Consider option --jobs-selection any
 # *ERROR* servant-client-core is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
 # *ERROR* servant-foreign is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
--- a/src/HaskellCI.hs
+++ b/src/HaskellCI.hs
@@ -136,18 +136,7 @@ travisFromConfigFile
     -> FilePath
     -> m ByteString
 travisFromConfigFile args opts path = do
-    gitconfig <- liftIO readGitConfig
-    cabalFiles <- getCabalFiles (optInputType' opts path) path
-    config' <- findConfigFile (optConfig opts)
-    let config = optConfigMorphism opts config'
-    pkgs <- T.mapM (configFromCabalFile config) cabalFiles
-    (ghcs, prj) <- case checkVersions (cfgTestedWith config) pkgs of
-        Right x     -> return x
-        Left []     -> putStrLnErr "panic: checkVersions failed without errors"
-        Left (e:es) -> putStrLnErrs (e :| es)
-
-    let prj' | cfgGhcHead config = over (mapped . field @"pkgJobs") (S.insert GHCHead) prj
-             | otherwise         = prj
+    (gitconfig, config, ghcs, prj') <- taskFromConfigFile args opts path
 
     ls <- genTravisFromConfigs args config gitconfig prj' ghcs
     patchTravis config ls
@@ -222,18 +211,7 @@ bashFromConfigFile
     -> FilePath
     -> m ByteString
 bashFromConfigFile args opts path = do
-    gitconfig <- liftIO readGitConfig
-    cabalFiles <- getCabalFiles (optInputType' opts path) path
-    config' <- findConfigFile (optConfig opts)
-    let config = optConfigMorphism opts config'
-    pkgs <- T.mapM (configFromCabalFile config) cabalFiles
-    (ghcs, prj) <- case checkVersions (cfgTestedWith config) pkgs of
-        Right x     -> return x
-        Left []     -> putStrLnErr "panic: checkVersions failed without errors"
-        Left (e:es) -> putStrLnErrs (e :| es)
-
-    let prj' | cfgGhcHead config = over (mapped . field @"pkgJobs") (S.insert GHCHead) prj
-             | otherwise         = prj
+    (gitconfig, config, ghcs, prj') <- taskFromConfigFile args opts path
 
     genBashFromConfigs args config gitconfig prj' ghcs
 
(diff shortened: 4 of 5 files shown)


fixtures/fail-versions.github
```diff
@@ -1,3 +1,4 @@
 # FAILURE
-# *ERROR* servant-client-core is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
-# *ERROR* servant-foreign is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
+# *ERROR* servant-client-core is missing tested-with annotations for: ghc-7.8.1, ghc-7.8.2, ghc-7.8.3, ghc-7.8.4
+# *ERROR* servant-foreign is missing tested-with annotations for: ghc-7.8.1, ghc-7.8.2, ghc-7.8.3, ghc-7.8.4
+(Consider option --jobs-selection any)
```
phadej, line 4: Don't write (edtit) files by hand.

Use `cabal run golden -- --accept` (the `--accept` will accept and write differences).

fixtures/fail-versions.travis
```diff
@@ -1,3 +1,4 @@
 # FAILURE
-# *ERROR* servant-client-core is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
-# *ERROR* servant-foreign is missing tested-with annotations for: ghc-7.8.1,ghc-7.8.2,ghc-7.8.3,ghc-7.8.4
+# *ERROR* servant-client-core is missing tested-with annotations for: ghc-7.8.1, ghc-7.8.2, ghc-7.8.3, ghc-7.8.4
+# *ERROR* servant-foreign is missing tested-with annotations for: ghc-7.8.1, ghc-7.8.2, ghc-7.8.3, ghc-7.8.4
+(Consider option --jobs-selection any)
```
phadej, line 4: You break (unspoken) invariant that errors are single lines.

Perfectly, *Consider...* text would be output with `putStrLnInfo`. That would be easier from call site of `checkVersionsInfo` (unfortunately you have to do it in three places)

test/Tests.hs
```diff
@@ -104,10 +104,12 @@ cabalGoldenTest name outRef act = goldenTest name readGolden (transform <$> act)
 
     upd (packData -> contents) = BS.writeFile outRef contents
 
-    cmp x y | x == y = return Nothing
-    cmp out1 out2 = return . Just . mconcat $
-        [ diff out1 out2
-        ]
+    cmp :: [String] -> [String] -> IO (Maybe String)
```
phadej, line 107: This change is unnecessary.

phadej (changes requested):

test/Tests.hs
andreasabel: Indeed, but so is a redex `concat [x]` instead of simply `x`.  I improved code comprehension here.

fixtures/fail-versions.github
andreasabel: A thanks, didn't know that trick.

test/Tests.hs
andreasabel: Do you want your redex back?

phadej: The redex is an artifact of recent change: https://github.com/haskell-CI/haskell-ci/commit/6725ab5c

Cleaning it up is welcome, but I would prefer it done separately, it's not related to `--jobs-selections` change.

fixtures/fail-versions.travis
andreasabel: Is it worthwhile to drill up the API for that?
* `checkVersionsUniform` seems to be the easiest place to add the "Consider" info to the error.
* One level up, in returning from `checkVersions`, I cannot attach this info, because the error strings could come from `checkVersionsAny`, in which case I should not attach the info.
* I could follow your suggestion to try to attach the info at all three call sites of `checkVersions`.  But there I would have to query the value of `(cfgTestedWith config)` a second time.  This would water down the abstraction `checkVersions`.
* If it is important to use `putStrLnInfo` instead of sticking the info at the end of the error message, then I'd rather suggest to return a more structured result from `checkVersions`.  Something like
```haskell
data CheckVersionsResult a b
  = CheckVersionsSucceeded 
    { compilerVersions :: S.Set CompilerVersion
    , cvProject :: Project a b Package
    }
  | CheckVersionsFailed
    { cvErrors :: [String]
    , cvInfos :: [String]
    }
```
How would you like that?

test/Tests.hs
andreasabel: I see.

andreasabel:
I reimplemented the feature according to our conversation above (https://github.com/haskell-CI/haskell-ci/pull/444#discussion_r547261409).

The flaw in the ointment is that for internal reasons *INFO* has to be printed before *ERROR* (because `putStrLnErrs` is including program termination via `exitFailure`).  

Now I hope to have met your ideas on how this should be implemented.

andreasabel:
Please merge before this bit-rots!

andreasabel:
@phadej: The status of this PR shows "changes requested" but I have addressed the change requests.  

And that already 6 weeks ago!

Please merge now.

phadej:
I won't merge this *now*. I do not like the demanding tone of your message, nor your email you sent me. You have no rights to demand anything. Internet & textual communication is difficult, but a bare "ping" would been enough to get my attention.

That said, now looking after some time at this, I'm not at all sure that suggesting

```
*INFO* Consider option --jobs-selection any
```

is the right thing to do. What if the problem might be out-of-sync `tested-with` definitions. How we can know?

andreasabel:
> I won't merge this _now_. I do not like the demanding tone of your message, nor your email you sent me. You have no rights to demand anything. 

Well, no, but general rules of courtesy apply also to the online world.  From my perspective, this was our interaction on this PR:

1. I make a PR that improves the UX (for newcomers).
2. You demand changes.
3. I implement the changes.
4. You disappear from the screen.

Judge for yourself.  Not a pleasant experience for a first-time contributor.  I'd say once you demand changes, you cannot simply say "no" anymore or just leave this to die.  If you don't like such UX improvements or you are not sure about it, it would be fair this at the beginning.  Time is a precious resource, for you and for me.

You hold the keys here, you are in the position of power, and with power comes responsibility...

> Internet & textual communication is difficult, but a bare "ping" would been enough to get my attention.

Er, there _were_ three pings with sufficient time to respond.  How should I know you are ever getting back to this?  I am hanging in the limbo here which isn't a pleasant state.

Anyway, in the future I will confine myself to weekly or biweekly `ping @phadej`, for up to six weeks, before getting upset.

So let's not brawl about this any longer.  I can also empathize with you as a project owner, with PR that are not in the right style or not "needed".  I also cannot claim  moral superiority... 😊 

> That said, now looking after some time at this, I'm not at all sure that suggesting
> 
> ```
> *INFO* Consider option --jobs-selection any
> ```
> 
> is the right thing to do. What if the problem might be out-of-sync `tested-with` definitions. How we can know?

Yes, it is merely a hint to the user, where they could look for the problem.  It could be, as you say, out-of-sync `tested-with` values, or it could be the missing flag.  In the former case, the course of action is quite clear to the user: revisit the `tested-with` fields.  However, if the user intended all the values of `tested-with` as they are, they might be at a loss, so the hint is useful.

It is like with the "did you forget an attachment" warnings, they can be to the point, they can be spam.  But it is easier to ignore a hint than to discover the information.

phadej:
Sorry. I'm upset about this issue, and cannot see proceed with it any time soon. I don't know how long it will take me to "cool".

phadej:
I guess this is a hard lesson for me to not review anything, before I'm sure I like an idea in the first place. In other words not looking at technical aspects before there is an agreement there is a problem, and the suggest solution is worthy.

Sorry for wasting your time.

andreasabel:
Acknowledged.

Let's cool down.  I am not married to this feature: personally, I know `--job-selection` now; one can wait whether there will be similar user questions/requests in the future.

Consider harvesting the refactorings from my PR (named datatype `CheckVersionsResult`, factoring out of `taskFromConfigFile`).

PR #461, opened 2021-01-06
GitHub Actions: include list of GHC versions in matrix

frasertweedale:
Alongside "include" include the list of GHC versions in the matrix.
That is, instead of:

  matrix:
    include:
      - ghc: 8.10.3
        allow-failures: false
      - ghc: 8.8.4
        allow-failures: false

haskell-ci now produces:

  matrix:
    ghc:
      - 8.10.3
      - 8.8.4
    include:
      - ghc: 8.10.3
        allow-failures: false
      - ghc: 8.8.4
        allow-failures: false

This approach reduces the work required to introduce another
"dimension" into the matrix.  As a concrete example, the /notmuch/
library must be tested against various versions of libnotmuch.  With
the "ghc" version list defined, adding the list of notmuch versions
to test yields a matrix that is the cartesian product of the two (or
more) lists.  The "include" dicts correctly propagate the
"allow-failures" fields to corresponding matrix legs (evidence:
https://github.com/purebred-mua/hs-notmuch/actions/runs/466042178).


fixtures/all-versions.github
```diff
@@ -25,6 +25,45 @@ jobs:
...
+          - 7.4.2
+          - 7.4.1
+          - 7.2.2
+          - 7.2.1
+          - 7.0.4
+          - 7.0.3
+          - 7.0.2
+          - 7.0.1
         include:
           - ghc: 8.10.3
```
phadej, line 68: How these would match against your proposed multi-dimension matrx?

Note: there is plan to change the matrix to be

```
- compiler: ghc
  version: 8.10.3
```

as we want to support GHCJS

frasertweedale: @phadej my objective is to enable addition of more matrix dimensions and associated build steps via `--github-patches`.  For example, see https://github.com/purebred-mua/hs-notmuch/commit/8b1d0f0cfac50e9df60ca39d9eccbb07d00d3f99#diff-486acc743e8f02cb435d1b164fe0a8389b3572f0199d79ed24c9fa7fe686f21b .  That patch works with this PR and, importantly, remains applicable and works as intended even as the GHC versions change, saving maintainer busywork.

The alternative is that after every change to `Tested-With` and subsequent `haskell-ci regenerate`, the maintainer has to manually reinstate all their changes to the matrix.  I have always had to do this with Travis, and I do not want to keep doing it with GitHub Actions  :)

Unfortunately, the suggested split of the `ghc` field into `compiler` and `version` fields, as you described above, does not support this approach.

However, if instead you could do it like:
```
hc:
  - ghc-8.10.3
  - ghc-8.8.4
include:
  - hc: ghc-8.10.3
    allow-failures: false
  - hc: ghc-8.8.4
    allow-failures: false
```

Then that *would* enable maintainers to add more matrix dimensions via a patch (and that patch can remain valid as `Tested-With` changes).

phadej: having `hc: ghc-8.8.4` would mean that we need to go to regexp magic to separate version from the compiler (which we had to do in Travis for other reasons). And we couldn't have `if: matrix.compiler == "ghcjs"` checks. 

I'm sorry, but this would mean a regression in code quality which I cannot accept now. You can maintain a fork of `haskell-ci`, for now.

frasertweedale: @phadej OK, how would you feel about a first-class `haskell-ci` option/argument to add the new matrix dimension, and perform the expansion within haskell-ci.  So something like:

`$ haskell-ci --expand-matrix "notmuch:0.31.3,git"`

To result in:

```
matrix:
  include:
    - ghc: 8.10.3
      notmuch: 0.31.3
      allow-failures: false
    - ghc: 8.10.3
      notmuch: git
      allow-failures: false
    - ghc: 8.8.4
      notmuch: 0.31.3
      allow-failures: false
    - ghc: 8.8.4
      notmuch: git
      allow-failures: false
```

And similarly under the planned change to have separate `compiler` and `version` fields.

Would this approach be acceptable?

frasertweedale: @phadej I submitted PR #462 implementing the alternative approach I suggested above.  Closing this PR.

frasertweedale:
Closed in favour of https://github.com/haskell-CI/haskell-ci/pull/462.

PR #473, opened 2021-02-08
GHA backend: Add head.hackage support

RyanGlScott:
This makes the GitHub Actions backend respect the `--head-hackage` flag, much in the same way that the Travis backend does.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -89,6 +89,7 @@ jobs:
           echo "HCNUMVER=$HCNUMVER" >> $GITHUB_ENV
           echo "ARG_TESTS=--enable-tests" >> $GITHUB_ENV
           echo "ARG_BENCH=--enable-benchmarks" >> $GITHUB_ENV
+          if [ $((HCNUMVER > 81003)) -ne 0 ] ; then echo "HEADHACKAGE=true" >> $GITHUB_ENV ; else echo "HEADHACKAGE=false" >> $GITHUB_ENV ; fi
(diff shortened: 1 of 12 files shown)


RyanGlScott:
As a proof of concept, I've demonstrated the use of `head.hackage` [here](https://github.com/RyanGlScott/gha-sandbox/runs/1851053123?check_suite_focus=true#step:16:20), a GitHub Actions job in which I specify `head-hackage >= 8.10`. With this, I am able to compile `critbit-0.2.0.0` on GHC 8.10, which crucially relies on a patch from `head.hackage`.

phadej:
Unrelatedly, why `critbit` relies on a patch for GHC-8.10 which is far from "new" GHC release?

phadej:
To answer myself, looks like it is a long ago abandoned package. I'd suggest changing that dependency)

RyanGlScott:
It is indeed worth wondering why `head.hackage` has a patch for `critbit` at all. For the sake of demonstrating that this patch works, however, `critbit` is actually quite convenient, as I needed an example of a library that won't build on GHC 8.10 without `head.hackage`, but would build with `head.hackage` added. (I need 8.10 in particular, as that is currently the most recent GHC version that the Ubuntu PPA supports.)

RyanGlScott:
Is this waiting on anything in particular before a review?

phadej:
No, it waits for me to have proper slot of time

phadej:
Rebased, amended a bit and merged in https://github.com/haskell-CI/haskell-ci/pull/477. Thanks

PR #484, opened 2021-02-21
Work around haskell/cabal#7291 by writing allow-newer to cabal.project, not ~/.cabal/config

RyanGlScott:
Due to haskell/cabal#7291, attempting to use `haskell-ci` with `head.hackage` currently will result in `cabal v2-update` hanging forever. I have worked around the issue by writing the `allow-newer` information to `cabal.project` rather than `~/.cabal/config`.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -119,7 +119,6 @@ jobs:
             url: http://hackage.haskell.org/
           EOF
           if $HEADHACKAGE; then
-          echo "allow-newer: $($HCPKG list --simple-output | sed -E 's/([a-zA-Z-]+)-[0-9.]+/*:\1/g')" >> $CABAL_CONFIG
           cat >> $CABAL_CONFIG <<EOF
           repository head.hackage.ghc.haskell.org
              url: https://ghc.gitlab.haskell.org/head.hackage/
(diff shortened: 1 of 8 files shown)


.github/workflows/haskell-ci.yml
```diff
@@ -118,9 +118,11 @@ jobs:
           repository hackage.haskell.org
             url: http://hackage.haskell.org/
           EOF
+          cat $CABAL_CONFIG
           if $HEADHACKAGE; then
-          echo "allow-newer: $($HCPKG list --simple-output | sed -E 's/([a-zA-Z-]+)-[0-9.]+/*:\1/g')" >> $CABAL_CONFIG
-          cat >> $CABAL_CONFIG <<EOF
+          touch cabal.project
+          echo "allow-newer: $($HCPKG list --simple-output | sed -E 's/([a-zA-Z-]+)-[0-9.]+/*:\1,/g')" >> cabal.project
```
phadej, line 124: we don't need this here. we don't need `allow-newer` to run `cabal v2-update` nor for `sdist`.

Add generation of `allow-newer` to the second `cabal.project` which is used with tests.

phadej (changes requested):

.github/workflows/haskell-ci.yml
phadej: To clarify, `allow-newer` were written to `cabal.config` because it was convenient. `repostory` stanza should stay in `cabal.config`.

RyanGlScott: Good suggestion. I've pushed a new version which implements this idea, which has the fortunate side effect of simplifying the implementation in the Travis backend.

phadej (approved):
Patch looks a lot cleaner now.

I'll merge once CI is green.

PR #485, opened 2021-02-22
Allow running cabal-docspec in constraint-sets

RyanGlScott:
See https://github.com/ekmett/ad/runs/1953010049?check_suite_focus=true#step:27:89 for an example of this in action.


phadej:
Thanks!

PR #488, opened 2021-02-23
[ Fix #487 ] enable bash-completion for file and directories

### LINKED ISSUE #487: bash-completion does not work properly
`haskell-ci` uses `optparse-applicative` which supports bash-completion.  I set up bash-completion for `haskell-ci` but while this allows me to complete commands, it breaks completion of files.  E.g.
```
$ haskell-ci git<TAB>
```
completes to
```
$ haskell-ci github 
```
but then 
```
$ haskell-ci github h<TAB>
```
is stuck (bell sound) even though the current directory contains `hackage-cli.cabal`.
Compare this to the spec:
```
$ haskell-ci github --help
Usage: haskell-ci github CABAL.FILE
  Generate GitHub Actions config

Available options:
  CABAL.FILE               Either <pkg.cabal> or cabal.project
```

I suspect that `action "file"` is missing in some places, see https://github.com/pcapriotti/optparse-applicative#actions-and-completers.

I'll give it a try (expect a PR).

andreasabel:
# Fix #487: enable bash-completion for file and directories

`optparse-applicative` supports bash-completion.  For this to work,
completers have to be supplied with option arguments, in particular

  - `action "file"` for FILE arguments
  - `action "directory"` for DIR arguments

This patch adds the necessary `action`s

  - directly to the option compositions (`O.strOption` ...)
  - via `metaActionHelp` to the options generated through the `OptparseGrammar`

This fixes bash-completion for haskell-ci, e.g.
```
  haskell-ci git<TAB>
  haskell-ci github
  haskell-ci github ha<TAB>
  haskell-ci github haskell-ci.cabal
  haskell-ci github haskell-ci.cabal --hl<TAB>
  haskell-ci github haskell-ci.cabal --hlint
  haskell-ci github haskell-ci.cabal --hlint-y<TAB>
  haskell-ci github haskell-ci.cabal --hlint-yaml
  haskell-ci github haskell-ci.cabal --hlint-yaml hl<TAB>
  haskell-ci github haskell-ci.cabal --hlint-yaml hlint.yaml
```

### DIFF
--- a/src/HaskellCI/OptparseGrammar.hs
+++ b/src/HaskellCI/OptparseGrammar.hs
@@ -25,14 +25,14 @@ import qualified Options.Applicative         as O
 import HaskellCI.OptionsGrammar
 
 data SomeParser s where
-    SP :: (Maybe String -> Maybe String -> O.Parser (s -> s)) -> SomeParser s
+    SP :: (Maybe MetaVar -> Maybe BashCompletionAction -> Maybe Help -> O.Parser (s -> s)) -> SomeParser s
 
 newtype OptparseGrammar s a = OG [SomeParser s]
   deriving Functor
 
 runOptparseGrammar :: OptparseGrammar s a -> O.Parser (s -> s)
 runOptparseGrammar (OG ps) = fmap (foldr (flip (.)) id) $ many $ asum
-    [ p Nothing Nothing
+    [ p Nothing Nothing Nothing
     | SP p <- ps
     ]
 
@@ -42,7 +42,7 @@ instance Applicative (OptparseGrammar s) where
 
 instance C.FieldGrammar ParsecPretty OptparseGrammar where
     blurFieldGrammar l (OG ps) = OG
-        [ SP $ \v h -> fmap (l C.#%~) (p v h)
+        [ SP $ \v a h -> fmap (l C.#%~) (p v a h)
         | SP p <- ps
         ]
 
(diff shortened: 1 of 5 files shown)


src/HaskellCI/OptparseGrammar.hs
```diff
@@ -25,14 +25,14 @@ import qualified Options.Applicative         as O
 import HaskellCI.OptionsGrammar
 
 data SomeParser s where
-    SP :: (Maybe String -> Maybe String -> O.Parser (s -> s)) -> SomeParser s
+    SP :: (Maybe MetaVar -> Maybe BashCompletionAction -> Maybe Help -> O.Parser (s -> s)) -> SomeParser s
```
phadej, line 28: Use `Maybe Completer` and `bashCompleter` to defined the `metaActionHelp`.

If someone feels like adding completers for other arguments, the support would be there. Also less `String`s.

phadej: Relatedly, does `bashCompleter` work with `zsh`? The `optparse-applicative` docs are sparse on the topic.

phadej:
Btw, what is the `[ ... ] ...` commit message practice. We don't have any for `haskell-ci` (nor dictate any), but your commits stand out, and I'm not sure it's a good thing.

src/HaskellCI/OptparseGrammar.hs
andreasabel: Good, switched to `Completer`.

I haven't used `zsh` before but dived into it and installed it.  Got completion with zsh to work, it is analogous to bash completion.  The completion script is generated by
```
$ haskell-ci --zsh-completion-script `which haskell-ci`
```
and (by convention) to be placed in a file `_haskell-ci` somewhere on zsh's `FPATH`.  That did it for me.

andreasabel:
> Is this some Agda project thing, and you are just so used to it? :)

Well, picked this habit up from @UlfNorell.  In the `[...]` there is a "theme", usually with some reference to an issue number, or a component.  

Remove it now in my commit here, placing the _theme_ before a colon.

phadej:
Amended and merged in https://github.com/haskell-CI/haskell-ci/pull/490, thanks!

andreasabel:
Thanks!

PR #496, opened 2021-03-11
GitHub Actions: GHCJS support

RyanGlScott:
I've tested this out on `singletons` [here](https://github.com/goldfirere/singletons/runs/2090801268), and it appears to work as you'd expect.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -45,7 +45,7 @@ jobs:
           nickname: github-actions
           server: irc.freenode.org
   linux:
-    name: Haskell-CI - Linux - GHC ${{ matrix.ghc }}
+    name: Haskell-CI - Linux - ${{ matrix.compiler }}
     runs-on: ubuntu-18.04
     container:
       image: buildpack-deps:bionic
@@ -53,17 +53,17 @@ jobs:
     strategy:
       matrix:
         include:
-          - ghc: 9.0.1
+          - compiler: ghc-9.0.1
             allow-failure: false
-          - ghc: 8.10.4
+          - compiler: ghc-8.10.4
             allow-failure: false
-          - ghc: 8.8.4
+          - compiler: ghc-8.8.4
             allow-failure: false
-          - ghc: 8.6.5
+          - compiler: ghc-8.6.5
             allow-failure: false
-          - ghc: 8.4.4
+          - compiler: ghc-8.4.4
             allow-failure: false
-          - ghc: 8.2.2
+          - compiler: ghc-8.2.2
             allow-failure: false
       fail-fast: false
     steps:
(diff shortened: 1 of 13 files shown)


phadej (approved):
This looks very simple! Thanks!

I'll try this on few repositories, and if it works there too, I'll merge and make a release.

Thanks again. This is great.

phadej:
I get errors

```
No haskell-ci.sh, skipping bash regeneration
*ERROR* ShellCheck! if echo $CC | grep -q ghcjs; then
SC1009: The mentioned syntax error was in this if expression.
SC1073: Couldn't parse this then clause. Fix to allow more checks.
SC1072: Expected a command. Fix any mentioned problems and try again.
```

when build with non GHC-9.0

phadej (changes requested):
ShellCheck doesn't seem to be happy.

RyanGlScott:
It turns out that this fails `ShellCheck`, but for some reason, `ShellCheck` doesn't run on GHC 9.0, which is what I used to test this locally. Let me think of a way to make `ShellCheck` happy...

phadej:
We have

```haskell
SC.csShellTypeOverride = Just SC.Sh
```

atm. We could change that to `Bash`, but the justification should be good. (also macOS on GHA - which we don't yet support, but in theory that is possible too) has `bash-3` which is ancient.

RyanGlScott:
I can convince `ShellCheck` that this is OK by turning the `if echo $CC ...` part into a single command, like so:

```diff
diff --git a/src/HaskellCI/GitHub.hs b/src/HaskellCI/GitHub.hs
index e91fa26..6dd9e0a 100644
--- a/src/HaskellCI/GitHub.hs
+++ b/src/HaskellCI/GitHub.hs
@@ -101,14 +101,15 @@ makeGitHub _argv config@Config {..} gitconfig prj jobs@JobVersions {..} = do
     steps <- sequence $ buildList $ do
         -- This have to be first, since the packages we install depend on
         -- whether we need GHCJS or not.
-        when anyGHCJS $ githubRun' "Set GHCJS environment variables" envEnv $ do
-            sh "if echo $CC | grep -q ghcjs; then"
-            tell_env "GHCJS" "true"
-            tell_env "GHCJSARITH" "1"
-            sh "else"
-            tell_env "GHCJS" "false"
-            tell_env "GHCJSARITH" "0"
-            sh "fi"
+        when anyGHCJS $ githubRun' "Set GHCJS environment variables" envEnv $ sh $ concat
+            [ "if echo $CC | grep -q ghcjs; then"
+            , "\n" ++ tell_env' "GHCJS" "true"
+            , "\n" ++ tell_env' "GHCJSARITH" "1"
+            , "\nelse"
+            , "\n" ++ tell_env' "GHCJS" "false"
+            , "\n" ++ tell_env' "GHCJSARITH" "0"
+            , "\nfi"
+            ]
 
         githubRun' "apt" envEnv $ do
             sh "apt-get update"
```

This doesn't change the output at all, only the internal data structure that `ShellCheck` examines. I've gone ahead and pushed this change.

PR #504, opened 2021-04-10
GHA: Support --distribution and --submodules

### LINKED ISSUE #503: GHA backend: support --submodules
As nice as `source-repository-package`s are, there are some situations where supporting `git` submodules would also be nice. `haskell-ci` has a `--submodules` flag for this purpose, but at the moment the GitHub Actions backend does not support it. At first, I thought supporting this flag would be as simple as adding `submodules: true` to the `checkout@v2` action, but when I tried this, I got [this error](https://github.com/RyanGlScott/gha-sandbox/runs/2289847432?check_suite_focus=true#step:10:32):

```
The repository will be downloaded using the GitHub REST API
To create a local Git repository instead, add Git 2.18 or higher to the PATH
Error: Input 'submodules' not supported when falling back to download using the GitHub REST API. To create a local Git repository instead, add Git 2.18 or higher to the PATH.
```

Darn. It looks like we might have to resort to a manual `git submodule update --init` to do this.

RyanGlScott:
The `checkout` action requires Git 2.18 or later to function properly, so this is only available with Ubuntu 20.04 (Focal) or later.

Fixes #503.

### DIFF
--- a/src/HaskellCI/GitHub.hs
+++ b/src/HaskellCI/GitHub.hs
@@ -36,6 +36,7 @@ import HaskellCI.Config.HLint
 import HaskellCI.Config.Installed
 import HaskellCI.Config.PackageScope
 import HaskellCI.Config.Ubuntu
+import HaskellCI.Config.Validity
 import HaskellCI.GitConfig
 import HaskellCI.GitHub.Yaml
 import HaskellCI.HeadHackage
@@ -100,6 +101,8 @@ makeGitHub _argv config@Config {..} gitconfig prj jobs@JobVersions {..} = do
             [ ("CC", "${{ matrix.compiler }}")
             ]
 
+    -- Validity checks
+    checkConfigValidity config jobs
     when (cfgSubmodules && cfgUbuntu < Focal) $
         throwErr $ ShError $ "Using submodules on the GitHub Actions backend requires "
                           ++ "Ubuntu 20.04 (Focal Fossa) or later."
--- a/src/HaskellCI/GitHub.hs
+++ b/src/HaskellCI/GitHub.hs
@@ -41,6 +41,7 @@ import HaskellCI.GitHub.Yaml
 import HaskellCI.HeadHackage
 import HaskellCI.Jobs
 import HaskellCI.List
+import HaskellCI.MonadErr
 import HaskellCI.Package
 import HaskellCI.Sh
 import HaskellCI.ShVersionRange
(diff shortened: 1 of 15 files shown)


src/HaskellCI/GitHub.hs
```diff
@@ -515,11 +526,11 @@ makeGitHub _argv config@Config {..} gitconfig prj jobs@JobVersions {..} = do
         , ghJobs = Map.fromList $ buildList $ do
             item (mainJobName, GitHubJob
                 { ghjName            = actionName ++ " - Linux - ${{ matrix.compiler }}"
-                , ghjRunsOn          = "ubuntu-18.04" -- TODO: use cfgUbuntu
+                , ghjRunsOn          = "ubuntu-" ++ ubuntuVerNum
```
phadej, line 529: `runs-on` is in fact irrelevant, as there is container. So I'd use `ubuntu-18.04` or `ubuntu-20.04`.

src/HaskellCI/GitHub.hs
```diff
@@ -98,6 +100,14 @@ makeGitHub _argv config@Config {..} gitconfig prj jobs@JobVersions {..} = do
             [ ("CC", "${{ matrix.compiler }}")
             ]
 
+    -- Validity checks
+    when (cfgSubmodules && cfgUbuntu < Focal) $
+        throwErr $ ShError $ "Using submodules on the GitHub Actions backend requires "
+                          ++ "Ubuntu 20.04 (Focal Fossa) or later."
+
+    when (anyGHCJS && cfgUbuntu > Bionic) $
```
phadej, line 108: This check is general, i.e. would apply to travis and bash backends too.

OTOH, the checkout nonsense is GHA specific.

src/HaskellCI/Travis.hs
```diff
@@ -93,6 +94,10 @@ makeTravis argv config@Config {..} prj jobs@JobVersions {..} = do
 
     -- before install: we set up the environment, install GHC/cabal on OSX
     beforeInstall <- runSh $ do
+        -- Validity checks
+        when (anyGHCJS && cfgUbuntu > Bionic) $
```
phadej, line 98: re. Yes, can we have common checks in one place, and not copy-pasted.

RyanGlScott: I have factored out this check into a `checkConfigValidity` function, which now lives in a new `HaskellCI.Config.Validity` module.

phadej:
Released in https://hackage.haskell.org/package/haskell-ci-0.12.1

PR #509, opened 2021-05-21
Use libera.chat for IRC notifications

RyanGlScott:


RyanGlScott:
Hm, that didn't quite [work as I had expected](https://github.com/haskell-CI/haskell-ci/pull/509/checks?check_run_id=2643699137#step:2:63):

```
< :molybdenum.libera.chat 442 github-actions #haskell-lens :You're not on that channel
```

It's not immediately clear to me what went wrong.

phadej:
```
 :molybdenum.libera.chat 477 github-actions #haskell-lens :Cannot join channel (+r) - you need to be identified with services
 ```
 
 is probably the culprit.

**Comment by RyanGlScott** [comment]
I wonder if this is because the freenode version of #haskll-lens didn't require identifying with a registered nick, but the libera.chat version does? If so, what is the the best way to set up an IRC bot like this one? Hm.

**Comment by phadej** [comment]
The `+r` was on freenode's `#haskell-lens` on and off (IIRC). With `+r` you prevent the spam, but also useful one from build bots. I'd take `+r` off, as I think that having a registered bot would need setting up a place to run it (to store its secrets), and I have no idea if anyone have done that already.

**Comment by RyanGlScott** [comment]
Yes, this is quite tricky. It would be possible to set up the `github-actions` bot such that it uses a password stored in [GitHub secrets](https://docs.github.com/en/actions/reference/encrypted-secrets). On the other hand, this would mean that every single repo which uses the `github-actions` bot would need to add such a secret to its repository settings, which is a much higher barrier to entry than before.

I wonder if @ekmett has any thoughts on this? He is one of the primary users of the IRC bot functionality, after all.

**Comment by phadej** [comment]
Any updates?

**Comment by RyanGlScott** [comment]
Not on my side. Unfortunately, I'm not quite sure how to proceed with this without some kind of coordination with someone who either (1) has the ability to manage permissions on the #haskell-irc channel, or (2) has admin permissions for repos that would make use of IRC bot functionality.

**Comment by ekmett** [comment]
I would really like a way to auth with services. Setting +r seems to be pretty critical in this current spam-bot era.

**Comment by RyanGlScott** [comment]
OK. I've been trying to figure out if there is an easier way to store an encrypted secret for the IRC bot's nick besides manually changing the settings for each repository which would make use of it. According to [this page](https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization), you can set up a secret that is shared among all repositories under a single GitHub organization. It's unclear to me though if it's possible to have a secret that is shared among all repositories by a particular user (e.g., https://github.com/ekmett).

**Comment by ekmett** [comment]
Worst case scenario I could make an organization and start moving my
repositories there. I could also suck it up and do per-repository secrets.

**Comment by phadej** [comment]
@RyanGlScott is there a bot + GHA integration which supports secrets, or should that still be built? If there isn't, I'd rather close this PR and wait until someone builds that bot first.

**Comment by RyanGlScott** [comment]
> is there a bot + GHA integration which supports secrets, or should that still be built?

I'm not sure if I understand your question correctly, but the phrase "supports secrets" isn't really meaningful. Secrets are a feature of GitHub Actions itself, so you can put `${{ secrets.MySecretValue }}` anywhere in a GHA YAML file. This is the mechanism that I would intend to use to supply an argument to [`irc-message-action`](https://github.com/marketplace/actions/irc-message-action), which is the IRC action that `haskell-ci` currently uses. `irc-message-action` has a [`sasl_password`](https://github.com/marketplace/actions/irc-message-action#sasl_password) parameter that lets you specify the password for an IRC bot account.

**Comment by phadej** [comment]
@RyanGlScott, so do I understand right:

1. someone, (perfectly owner of the repository?) should create a bot, register it on the libera network, store the secrets in their keychain.
2. `haskell-ci` should support supplying a `sasl_password` to the `irc-message-action`?
3. the person knowing the bot password should configure the GHA so the secret is there
4/ profit

If my understanding is right, then only 2. step requires changes to `haskell-ci`. Please change this PR to do that then.

The `haskell-ci` itself isn't using IRC notifications anymore so in that regard this PR is outdated.

**Comment by RyanGlScott** [comment]
> do I understand right

Yes, that's right.

> If my understanding is right, then only 2. step requires changes to `haskell-ci`. Please change this PR to do that then.

Will do.

**Comment by RyanGlScott** [comment]
I've pushed a new version that adds password functionality. I haven't had a chance to actually test this in the wild yet.

**Comment by phadej** [comment]
This looks fine. I'll merge it if you are not planning to test it *very soon* (so this PR doesn't bitrot).

**Comment by RyanGlScott** [comment]
I tested this (with minor modifications to use my own IRC bot) in [this commit](https://github.com/RyanGlScott/gha-sandbox/commit/349ca8191dfd70695b2f73febc2cf68b508364e7), and everything works as expected.

**Comment by RyanGlScott** [comment]
I suppose this raises the question: should we allow users to name the nick something besides `github-actions`?

**Comment by phadej** [comment]
Yes, I guess it will be required, or at least useful.

**Comment by RyanGlScott** [comment]
Alright, I've added an additional `--irc-nickname` option for maximal configurability.

File: src/HaskellCI/Config.hs [hunk 1]
```diff
@@ -178,7 +180,11 @@ configGrammar = Config
     <*> C.monoidalFieldAla    "branches"                  (C.alaList' C.FSep C.Token')        (field @"cfgOnlyBranches")
         ^^^ metahelp "BRANCH" "Enable builds only for specific branches"
     <*> C.monoidalFieldAla    "irc-channels"              (C.alaList' C.FSep C.Token')        (field @"cfgIrcChannels")
-        ^^^ metahelp "IRC" "Enable IRC notifications to given channel (e.g. 'irc.freenode.org#haskell-lens')"
+        ^^^ metahelp "IRC" "Enable IRC notifications to given channel (e.g. 'irc.libera.chat#haskell-lens')"
+    <*> C.freeTextField       "irc-nickname"                                                  (field @"cfgIrcNickname")
+        ^^^ metahelp "PASSWORD" "Nickname with which to authenticate to an IRC server. Only used if `irc-channels` are set."
```
phadej, line 185: `s/PASSWORD/NICKNAME/`

RyanGlScott: Done.

phadej:
Thanks!

RyanGlScott:
Unfortunately, it looks like even existing IRC configurations using Freenode are now broken—see [this example](https://github.com/ekmett/bound/runs/3092863108). I'm posting this here not because there is any concrete action that needs to be taken on `haskell-ci`'s end, but rather as a PSA to those who actively use `haskell-ci`'s IRC feature (most of whom are cc'd on this issue). In the meantime, I'll need to disable the Freenode-based IRC configurations as a temporary CI fix until we figure out a more permanent solution.

PR #512, opened 2021-05-27
Setup method

phadej:
Aiming for something like in https://github.com/haskell-CI/haskell-ci/issues/474#issuecomment-781652095, cc @RyanGlScott


phadej:
I have tested ghcup install method also in

- https://github.com/haskellari/splitmix/pull/56
- https://github.com/well-typed/optics/pull/422

so I'm gonna merge this. Please try out and report if works or doesn't :)

PR #521, opened 2021-06-06
Add 8.10.5 to known compilers

phadej:


andreasabel:
Nice.

It seems though, that ghc-8.10.5 isn't in hvr:ppa yet: https://github.com/BNFC/bnfc/runs/2804114742?check_suite_focus=true

andreasabel:
Still a problem as of today:
```
E: Unable to locate package ghc-8.10.5
E: Couldn't find any package by glob 'ghc-8.10.5'
E: Couldn't find any package by regex 'ghc-8.10.5'
```

PR #524, opened 2021-06-16
Remove uses of allow-newer from cabal.project

RyanGlScott:
The latest Hackage versions of these libraries all support GHC 9.0.


phadej:
Thanks!

Issue #395, opened 2020-05-31
[Solution included] HCPKG seems to be set wrongly for osx


Mikolaj:
Here:

https://travis-ci.org/github/LambdaHack/LambdaHack/jobs/693252187#L206

if fails with

/Users/travis/.travis/functions: line 109: /Users/travis/.ghc-pkg-install/ghc/bin/ghc-8.4.4: No such file or directory

The same for other GHCs.

Mikolaj:
The line defining osx stuff should be change to the following and then it works fine:

  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then HC=$HOME/.ghc-install/ghc/bin/$TRAVIS_COMPILER; WITHCOMPILER="-w $HC"; HCPKG=$HOME/.ghc-install/ghc/bin/${TRAVIS_COMPILER/ghc/ghc-pkg}; CABAL=$HOME/.ghc-install/ghc/bin/cabal; fi

phadej:
https://hackage.haskell.org/package/haskell-ci-0.10.2

Mikolaj:
Thank you!

PR #528, opened 2021-08-02
Fix a typo in the fix for #395.

merijn:
The previous fix for #395 causes ShellCheck to barf and seems to be a typo.

See https://github.com/haskell-CI/haskell-ci/issues/395


phadej:
Thanks

PR #529, opened 2021-08-02
Allow diffing of config files and regendata

merijn:
As I talked about *ages* ago on IRC, I finally got around to finishing it up.

Adds a new `diff-config` command that takes a config file and/or generated CI workflow and prints a diff of the configurations. There's no tests atm and there is some bikeshedding to be done with regards to: Should we take commandline flags/existing config files into account? If yes, should those be applied to the "old" or the "new" config?

In principle this also obsoletes/can replace the old dump-config command.


merijn:
@phadej How do you feel about this? Obviously needs a bit more polish in the form tests and bike-shedding how commandline flags affect the diff command (if they even should?). The current Dump functionality can be trivially replaced by (and augmented to output dumps arbitrary configs, instead of just the default).

I'm just to lazy to finish up the last polish if you're not interested anyway :p

PR #534, opened 2021-08-24
Respect CABAL_DIR in 'resolveConfig'

hasufell:
Although 'findConfig' already respects it, 'resolveConfig' will
try to fill in missing values by examining 'getAppUserDataDirectory',
ignoring CABAL_DIR and possibly picking a wrong path for 'cfgStoreDir'.

This causes failures in e.g. cabal-docspec: https://gitlab.haskell.org/haskell/ghcup-hs/-/jobs/770892#L2501

----

It should also be backported to pre-Cabal-3.6, so we can build a `cabal-docspec` that includes it.

### DIFF
--- a/cabal-install-parsers/src/Cabal/Config.hs
+++ b/cabal-install-parsers/src/Cabal/Config.hs
@@ -74,9 +74,14 @@ findConfig = do
     case env of
         Just p -> return p
         Nothing -> do
-            cabalDirVar <- lookupEnv "CABAL_DIR"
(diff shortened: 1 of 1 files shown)


hasufell:
Can be tested with:

* unpatched: https://downloads.haskell.org/~ghcup/unofficial-bindists/cabal-docspec/cabal-docspec-0.0.0.20210228.tar.bz2
* patched: https://downloads.haskell.org/~ghcup/unofficial-bindists/cabal-docspec/cabal-docspec-0.0.0.20210228_p1.tar.bz2

PR #546, opened 2021-10-09
cabal-install-parsers: Allow to build with aeson >=2

hasufell:
Tests pass

----

Also maybe do a revision update on hackage?


phadej:
Something is holding `aeson-2` from being used in CI.

Issue #553, opened 2021-11-10
haskell-ci jobs sometimes hangs


phadej:
I remember @RyanGlScott reporting this. By looking at `runner` issue tracker, it seems there are similar-ish reports for non-Haskell jobs too:
- https://github.com/actions/runner/issues/1473
- https://github.com/actions/runner/issues/1326

I don't know whether we can do anything, as there are no clues in the issues.

phadej:
I added a default 60min timeout so at least the hanging job hang themselves quicker. The memory limit didn't help.

RyanGlScott:
Thanks!

Out of curiosity, how does one determine what memory limit to use? I noticed that the Travis backend and the GitHub Actions backend have different memory limits ([6G](https://github.com/haskell-CI/haskell-ci/blob/a5ba1674df06ddd4a31af752af26c23fb645a639/src/HaskellCI/Travis.hs#L202) versus [3G](https://github.com/haskell-CI/haskell-ci/blob/4d24492e0a7111e43780548c592046b42f6709c2/src/HaskellCI/GitHub.hs#L285), respectively).

phadej:
virtual environments on GHA has 7G memory (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners), because dependencies are -j2 (i.e. two GHCs) 3G is conservative.  6G for travis was optimistic, but there we knew for sure when job failed due no memory available, in GHA i'm just guessing

phadej:
I'm trying to give a memory hug to CI in #556, but looks like big record is not working.

phadej:
This issue is resolved. See discussion in https://github.com/haskell-infra/hackage-trustees/issues/319

PR #554, opened 2021-11-11
ghc-options with memory limit

phadej:


phadej:
I tried this hoping it would solve #553, but the problem seems to persist (in `aeson`).

PR #539, opened 2021-09-01
cabal-install-parsers: Construct wildcard glob directly

RyanGlScott:
Rather than relying on `FilePathGlob`'s `ReadP`-based parser, which behaves erratically on Windows (see #478), we now take a cue from `cabal-install` (see https://github.com/haskell/cabal/blob/d8c191d6c574d8295696e1e9b32ca16bd3cc3f5c/cabal-install/src/Distribution/Client/ProjectConfig.hs#L896-L946) and construct a glob for `./*.cabal` directly.

Fixes #478.

PR #558, opened 2021-11-13
Revert 539

phadej:
This kinds of reverts the #539 patch, but then reapplies a slight modification of it.


RyanGlScott (approved):
I tried this out on Windows, and as far as I can tell, it works as expected. All of the `haskell-ci` tests pass, and I am successfully able to run `haskell-ci regenerate` on both `haskell-ci`'s `cabal.project` file as well as another project that simply uses `packages: .`.

Thanks for fixing my mistake!

PR #563, opened 2021-11-29
Add sourcehut generator

fgaz:


phadej:
Don't edit readme.

src/HaskellCI/Cli.hs
```diff
@@ -23,6 +25,7 @@ data Command
     = CommandTravis FilePath
     | CommandBash FilePath
     | CommandGitHub FilePath
+    | CommandSourcehut (SourcehutOptions (Maybe String))
```
phadej, line 28: Don't introduce separate SourceHut options. Everything is configured through `Config`.

fgaz: Yeah I was about to ask about this since it only seems to introduce problems (especially parseOptions). Will do.

fgaz:
Hmm, why isn't CI running?

src/HaskellCI/Cli.hs
fgaz: Done

fixtures/copy-fields-some.github
```diff
@@ -2,15 +2,15 @@
...
+#   haskell-ci '--copy-fields=some' 'github' 'copy-fields-some.project' '--sourcehut-source' 'https://example.org'
 #
 # To regenerate the script (for example after adjusting tested-with) run
 #
 #   haskell-ci regenerate
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# REGENDATA ["--copy-fields=some","github","copy-fields-some.project"]
+# REGENDATA ["--copy-fields=some","github","copy-fields-some.project","--sourcehut-source","https://example.org"]
```
phadej, line 13: Why these are in non sourcehut outputs?

test/Tests.hs
```diff
@@ -70,7 +81,7 @@ fixtureGoldenTest fp = testGroup fp
         makeFlags :: IO ([String], Options)
         makeFlags = do
             argv <- readArgv
-            let argv' = argv ++ [name, projectfp]
+            let argv' = argv ++ [name, projectfp] ++ ["--sourcehut-source", "https://example.org"]
```
phadej, line 84: ???

fgaz: `--sourcehut-source` is required if a git remote is not found. Yeah, I should probably add that flag only if name=="sourcehut"

fgaz:
What should I do with the `tell_env` hint? The github backend also calls it like that, but it's inside a `where` there, which is fine for `hlint`

phadej:
> What should I do with the tell_env hint? The github backend also calls it like that, but it's inside a where there, which is fine for hlint

I don't know. `hlint` is probably drunk. Add ignore to `.hlint.yaml`, I don't like "Use camelCase" hint anyway.

test/Tests.hs
fgaz: done

phadej:
@fgaz what is the status of this PR?

fgaz:
I've been using this branch for my personal projects. I don't remember too well what was the status of this but it looks like all comments are resolved. Let's rebase and see if tests still pass

fgaz:
Hmm, I guess I'll need to update it to use ghcup since the ppa is no longer maintained

fgaz:
Here you go, this should work

ysangkok:
It's too bad that this now has conflicts, I wonder how hard they would be to resolve. I personally think it's great that you did this work @fgaz , and I still think it would be a great addition. Are you still using it?

PR #566, opened 2021-12-03
Bump doctest default & hardcoded versions for GHC 9.2

raehik:
The current default doctest version doesn't work with GHC 9.2. The latest version supports all the GHCs we need, so I've updated the default version and everywhere that a doctest constraint is formed.


raehik:
Targets #565 .

PR #582, opened 2022-04-16
Remove `hasLibrary` guard from haddock step

andreasabel:
Remove `hasLibrary` guard from haddock step

Rationale:
- `cabal v2-haddock` also succeeds for `executable` packages
- one can always opt-out with `haddock: False` in `cabal.haskell-ci`

In the wild: 
- https://github.com/haskell/hackage-server/pull/1061


andreasabel:
@phadej, does this change sound reasonable?

phadej:
Does this actually work?

With:

```cabal
cabal-version: 2.4
name: pkkg
version: 0

executable foo
  default-language: Haskell2010
  main-is: exe.hs
  build-depends: base
```

```haskell
module Main (main, foo) where

foo :: Int
foo = 42

main :: IO ()
main = print foo
```

and

```console
% cabal --version
cabal-install version 3.6.2.0
compiled using version 3.6.2.0 of the Cabal library 
```

trying to `cabal haddock all` fails:

```console
% cabal haddock all
cabal: renderBuildTargetProblem: unexpected status (TargetDisabledByUser,Nothing)
CallStack (from HasCallStack):
  error, called at src/Distribution/Client/CmdErrorMessages.hs:356:14 in main:Distribution.Client.CmdErrorMessages
```

I tried to find an issue in `Cabal` about this behavior, but could find it.

andreasabel:
> Does this actually work?

Seems like for your example, no haddock is actually run.

> I tried to find an issue in `cabal-install` about this behavior, but could find it. Is it fixed recently?

I can confirm that the error is gone on `cabal-install` `master`.

Seems like one has to pass `--haddock-executables` to silence the warning:
```
No documentation was generated as this package does not contain a library.
Perhaps you want to use the --haddock-executables, --haddock-tests,
--haddock-benchmarks or --haddock-internal flags.
```

Let me try to investigate further... (I remembered it worked fine on `hackage-server`, but I have to revisit the details).

phadej:
`hackage-server` probably works cause it has internal libraries. But if we can build haddocks with `--haddock-all` to build haddocks for exes, tests and benchmarks, that will be great. These should generally succeed, there is no reason to have invalid markup in the auxiliary components.

andreasabel:
@phadej wrote:
> I tried to find an issue in `cabal-install` about this behavior, but could find it. Is it fixed recently?

Looks like it is still open:
- https://github.com/haskell/cabal/issues/5890

andreasabel:
> `hackage-server` probably works cause it has internal libraries. 

Yes, reading up on https://github.com/haskell/cabal/issues/5890 confirms this: if you have `other-modules`, docs are generated.  Thus, your example just hit the case that doesn't work.

> But if we can build haddocks with `--haddock-all` to build haddocks for exes, tests and benchmarks, that will be great.

Since `cabal v2-haddock --haddock-all` does not fail (even if it does not produce any HTML),  unless there is invalid markup, it might be no harm to run it by default, as you said.

andreasabel:
A few successful CI runs:
- https://github.com/andreasabel/cabal-clean/runs/6249085026?check_suite_focus=true
- https://github.com/andreasabel/agda2lagda/actions/runs/2254675955
- https://github.com/andreasabel/risc386/actions/runs/2254519652

phadej:
Merged in https://github.com/haskell-CI/haskell-ci/pull/585 (please don't edit version in the future if it's release i.e. even version).

andreasabel:
> Merged in #585 

Thanks!

> (please don't edit version in the future if it's release i.e. even version).

Ok, understood.

PR #591, opened 2022-05-29
Add 9.2.3 to known GHC versions

andreasabel:
Add 9.2.3 to known GHC versions
Tested e.g. here: https://github.com/agda/fix-whitespace/runs/6644545666?check_suite_focus=true

~~This PR still lacks an update of the fixtures.~~

Bumped version to 0.15.20220530 (I hope that's ok, it is a dev version).
Self-test with GHC 9.2.3 now (instead of GHC 9.2.2.).


phadej:
Thanks, merged in https://github.com/haskell-CI/haskell-ci/pull/593

PR #592, opened 2022-05-30
Don't unconditionally force allow-failure with head.hackage

arybczak:
I was surprised that it is like this since there's `allow-failures` for configuring which versions you want to allow failures for.

Generally, once the program compiles with pre-released GHC, I'd expect it to continue compiling so it's nice to be able to say that you want to see the failure if it occurs.


phadej:
AFAIK, there is no promise that `head.hackage` would be "production" ready. In the past, new releases (without patches in `head.hackage`) made CI fail after it once succeeded.

arybczak:
Uh. Can we at least make it accept failures by default, but be overridable?

PR #602, opened 2022-07-29
Bump github/actions to latest: checkout@v3, cache@v3, irc-message-action@v2

andreasabel:
Bump github/actions to latest: checkout@v3, cache@v3, irc-message-action@v2

I didn't change the  package `version` in `haskell-ci.cabal`.

### DIFF
--- a/fixtures/all-versions.github
+++ b/fixtures/all-versions.github
@@ -436,7 +436,7 @@ jobs:
           chmod a+x $HOME/.cabal/bin/cabal-plan
           cabal-plan --version
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           path: source
       - name: initial cabal.project for sdist
@@ -475,7 +475,7 @@ jobs:
           $CABAL v2-build $ARG_COMPILER $ARG_TESTS $ARG_BENCH --dry-run all
           cabal-plan
       - name: cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           key: ${{ runner.os }}-${{ matrix.compiler }}-${{ github.sha }}
           path: ~/.cabal/store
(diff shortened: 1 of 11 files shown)


fixtures/all-versions.github
```diff
@@ -406,7 +406,7 @@ jobs:
           chmod a+x $HOME/.cabal/bin/cabal-plan
           cabal-plan --version
       - name: checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
phadej, line 409: this action uses node-16 as per changelog. Does old settinngs for `distribution` work with it?

andreasabel: Frankly, I don't know whether this works with `distribution`.
In the long run, we will have to upgrade as `node-12` is deprecated.
Also, `dependabot` is harrassing me about outdated actions:
- https://github.com/agda/fix-whitespace/pull/39

lf-:
There is a reason to upgrade: This is in my logs:

```
Run actions/cache@v2
/usr/bin/docker exec  6a50ec8666c3927a0278b3a464bd654829d45470b87caff8207d2142c95119c4 sh -c "cat /etc/*release | grep ^ID"
Warning: The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
Cache not found for input keys: Linux-ghc-9.4.2-c06f4d5fd953f688c14963fad0aaf89e69553702, Linux-ghc-9.4.2-
```

haskell-ci pipelines will be broken by GitHub in late May 2023 due to the deprecation of functionality used in cache@v2.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

PR #613, opened 2022-08-22
Configurable haddock components

phadej:


src/HaskellCI/Config.hs
```diff
@@ -170,6 +173,8 @@ configGrammar = Config
         ^^^ metahelp "RANGE" "Build benchmarks"
     <*> rangeField           "haddock"                                                        (field @"cfgHaddock") anyVersion
         ^^^ metahelp "RANGE" "Haddock step"
+    <*> C.optionalFieldDef   "haddock-components"                                             (field @"cfgHaddockComponents") ComponentsAll
```
andreasabel, line 176: @phadej: Ah, I see that `ComponentsAll` is the default.  This is backwards compatible to my `--haddock-all` PR, thanks.  But I come to think it would be better to be backwards compatible to the latest release of Haskell-CI (0.14.3?).  This would suggest that the default should be `ComponentsLibs`.

phadej: General philosophy of `haskell-ci` is to have as much checks enabled by default. If new checks break your build, other disable them or fix your code.

(Enabling new checks almost everywhere is the busy work I want to avoid).

PR #615, opened 2022-09-01
Adding ability to specify mulitple postgresql versions to be used in builds

kubek2k:
this adds a new parameter `postgresql-versions` that one can use to specify more than one postgresql version that is later used in matrix.

### DIFF
--- a/src/HaskellCI/Travis.hs
+++ b/src/HaskellCI/Travis.hs
@@ -454,13 +454,13 @@ makeTravis argv config@Config {..} prj jobs@JobVersions {..} = do
             when cfgPostgres $ item "postgresql"
         , travisAddons        = TravisAddons
             { taApt          = TravisApt [] []
-            , taPostgres     = if cfgPostgres then Just "10" else Nothing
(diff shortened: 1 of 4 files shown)


src/HaskellCI/Travis.hs
```diff
@@ -494,13 +494,17 @@ makeTravis argv config@Config {..} prj jobs@JobVersions {..} = do
...
                                 , taGoogleChrome = False
                                 }
                             }
-
-                for_ (reverse $ S.toList linuxVersions) $ tellJob False
-                for_ (reverse $ S.toList macosVersions) $ tellJob True
+                let postgresVersions = if cfgPostgres 
+                                          then Just <$> cfgPostgresVersions
+                                          else [Nothing]
+                for_ (reverse $ S.toList linuxVersions) $ \v ->
```
arybczak, line 504: This can be `toDescList`. Same below.

arybczak:
@phadej could this be merged? I wanted to add an option to configure PostgreSQL version used (since right now 10 is hardcoded), but this is much better.

PR #620, opened 2022-09-21
Add Nix flake

lf-:
I've added a Nix flake to allow running haskell-ci in one command from any
computer with Nix. I've tested this on an aarch64 Mac.

I've updated the README to match as well.

Care and feeding of nix flakes:

There's a lockfile at `flake.lock` containing all the inputs including the
nixpkgs set. The overrides I've put in the overlay are due to the stackage LTS
that nixpkgs is currently based on being slightly old. They will error on
evaluation if they become unnecessary following a lockfile change.

You can update the nixpkgs version like so:

```
nix flake lock --update-input nixpkgs
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,31 +8,52 @@ There is also legacy Travis-CI configuration generator, which is unmaintained.
 or [`ghcup`](https://www.haskell.org/ghcup/) to install GHC
 and `cabal-install`.
 
-GHC-7.0.1 &mdash; GHC-9.2.1 are supported.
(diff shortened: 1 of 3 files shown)


phadej (changes requested):
I don't use nix. I also know it so poorly, I cannot fix issues if these arises.

For me `ghcup install ghc` and `cabal` and then `cabal build` work. I don't see a value in bringing in Nix, I'm sorry.

PR #633, opened 2023-01-17
Save GitHub cache on failure as well

maksbotan:
The cache contains built dependencies. When job fails due to compilation error in the project itself, or due to test failure, next one will have to recompile the same dependencies all over again, because they were not cached for failed job.

This commit saves cache unconditionally, using recently introduced GitHub action 'actions/cache/save@v3'.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -148,7 +148,7 @@ jobs:
         run: |
           $CABAL v2-update -v
       - name: cache (tools)
-        uses: actions/cache@v2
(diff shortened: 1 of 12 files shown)


maksbotan:
Sample job which restored cache after failure: https://github.com/maksbotan/haskell-ci/actions/runs/3939978470/jobs/6740780009

maksbotan:
@phadej hi! can you please take a look at this one?

phadej:
Great. I was wishing for this functionality in GHA cache for some time!

maksbotan:
Thank you!

PR #646, opened 2023-03-10
Add -Werror=missing-fields

phadej:


ulysses4ever:
[Unlike unused packages](https://github.com/haskell-CI/haskell-ci/pull/789#issuecomment-3224566698), there's no way to turn this off, is there? This is too harsh...

ulysses4ever:
I can't easily fix a code that have hundreds of these. 

Missing methods option isn't the same as missing fields, is it?

phadej:
Which makes  me think that I also should add an option for `-Werror=orphans`, as all "good" code (which unfortunately has to defined orphan instances) cares to disable that warning in modules with orphans.

PR #648, opened 2023-03-12
Update GHC 9.6.1 to point to the released GHC

andreasabel:
- Update GHC 9.6.1 to point to the released GHC
- bump `ghcup` to latest
- bump `cabal` to latest


andreasabel:
Ah this PR is redundant, I guess, because:
- #647

PR #650, opened 2023-03-20
Bump major version of doctest to 0.21

mimi1vx:
Major change for doctest is support of new cmdline options and support for ghc-9.6 so is safe to bump to new major.


phadej:
Thanks. Merged as part of https://github.com/haskell-CI/haskell-ci/pull/651

PR #666, opened 2023-07-07
Use HLint 3.6.1

phadej:


RyanGlScott:
One thing that I discovered when trying out the `hlint-3.6.1` bindist is that dynamically links against the `glibc` version provided by Ubuntu 22.04 (Jammy). This will _not_ work if you run `hlint-3.6.1` on Ubuntu 20.04 (Bionic), which is the default setting for `haskell-ci`. [An example of this failing](https://github.com/ekmett/lens/actions/runs/5481698154/jobs/9993871837#step:12:41):

```
hlint: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by hlint)
hlint: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by hlint)
```

I'm not sure if there is a good way to handle this on the `haskell-ci` side. Perhaps a disclaimer in the documentation somewhere?

phadej:
> I'm not sure if there is a good way to handle this on the haskell-ci side.

There isn't.

> Perhaps a disclaimer in the documentation somewhere?

Nobody reads docs, until there is problem, then they google and will find this issue.

---

I opened an upstream issue: https://github.com/ndmitchell/hlint/issues/1533, and will withhold merging this PR until it's resolved in a way or another.

marinelli:
It's possible to specify, for a particular GitHub _Job_, a different image [1]. So you could use by default the ubuntu-20 image, but for the _linting_ jobs use ubuntu-22. Anyway, I agree that a statically linked binary is a much more resilient solution.

[1] https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

RyanGlScott:
This is likely obsolete after https://github.com/haskell-CI/haskell-ci/pull/717.

PR #669, opened 2023-07-10
Add `apt` field to template `cabal.haskell-ci`

byorgey:
The `apt` flag is accepted on the command line and in the `cabal.haskell-ci` config file, but it is not documented anywhere making it difficult to discover.


phadej:
Thanks

PR #689, opened 2023-11-19
Change branch name in README.md

jcranch:
haskell-ci creates a script that runs on branches called `main` and `ci-*`; however, `README.md` advised creating and pushing a branch called `new-ci`.

Of course one could change the code to match the documentation instead of changing the documentation to match the code.


phadej:
Thanks for spotting this.

I removed the whole step, as it's not worth it telling people how to use GitHub. They probably know how.

Issue #690, opened 2023-11-23
Use latest `doctest` by default


andreasabel:
I suggest to remove the default constraint on `doctest` so that always the latest version gets installed unless the user asks for a specific version.
https://github.com/haskell-CI/haskell-ci/blob/e0dd023ac1eff2b40bd31df1184c5bf7ed9ae176/src/HaskellCI/Config/Doctest.hs#L28-L29
`doctest` has seen frequent major releases recently, and the current bound is already outdated.  Even if updated, it will rot quickly.

phadej:
The latest `doctest` should be "standalone"  run as `cabal repl --with-ghc=doctest`, as I'm not using `doctest` myself anymore, I haven't made according changes to the CI generation either.

andreasabel:
> @phadej locked the conversation on #691.

Apologies, but I don't suffer to be muzzled.

@phadej wrote: 
> as I'm not using `doctest` myself anymore, I haven't made according changes to the CI generation either.

From the interaction in #691 and from accumulated interactions on various other PRs, I am getting the impression that you consider `haskell-ci` as your private tool rather than trying to maximise its usability for the greater Haskell community.

I don't see how this continues Herbert's legacy.

I am beginning to believe our interaction on this project cannot be healed.  You do not seem to want any collaboration.

PR #691, opened 2023-11-23
By default, do not constrain `doctest`

andreasabel:
`doctest` bound has been rotting too quickly.
Just use the latest version by default.
Can always be overwritting by `doctest-version` option in `cabal.haskell-ci`.

Closes https://github.com/haskell-CI/haskell-ci/issues/690

### DIFF
--- a/src/HaskellCI/Config/Doctest.hs
+++ b/src/HaskellCI/Config/Doctest.hs
@@ -5,8 +5,6 @@ module HaskellCI.Config.Doctest where
 
 import HaskellCI.Prelude
 
-import Distribution.Version (majorBoundVersion)
-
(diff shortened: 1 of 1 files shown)


andreasabel:
> You can specify `doctest-version: >=0` in your configuration file.

Yes of course.
But good tooling has sensible defaults.  Defaulting to an outdated version isn't a good default.

andreasabel:
> Do we? If they worked, they should continue to work.

If they weren't broken, I wouldn't have gone to the effort of changing Haskell CI.
Let me see if I can recover the evidence...

phadej:
> If they weren't broken,

You added GHC-9.8 CI, so you did change *something*.

PR #697, opened 2023-12-24
Add regenerate hook for pre-commit.

alunduil:
This allows pre-commit to use haskell-ci to regenerate configuration as part of the pre-commit hook or CI with minimal setup per environment.

Tested with `pre-commit try-repo --all-files .`

Let me know if any documentation would be helpful to include with this setup or if there is a preference for a different configuration.  pre-commit isn't often used in the Haskell community but I've been using it to keep environments consistent for some time and would like it to have more Haskell tooling.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,7 +2,7 @@
   name: haskell-ci regenerate
   entry: haskell-ci regenerate
   language: haskell
-  files: '.*\.cabal'
+  files: '(\.cabal$)|(^.github/workflows/haskell-ci.yml$)|(.travis.yml$)'
   pass_filenames: false
   description: |-
     Regenerate Haskell CI configurations with haskell-ci.  Configuration must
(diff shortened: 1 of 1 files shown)


alunduil:
This is an addition of a hook specification for pre-commit (see https://pre-commit.com/#new-hooks) that allows users of pre-commit to directly use the tagged versions of haskell-ci in their pre-commit configurations with the following additions:

```yaml
repos:
  - repo: https://github.com/haskell-CI/haskell-ci
    rev: v0.16.6
    hooks:
      - id: regenerate
```

This small configuration ensures that haskell-ci configurations are regenerated upon git commit and, if configured, on all pull requests for the project.

pre-commit is a git hooks management system that has ready to use GitHub action configurations and does environment isolation and caching to ensure more repeatable processes.  Details can be found at https://pre-commit.com/ .

Please let me know if you have any other questions.

phadej:
Ok, so if I understand right, that bit of configuration doesn't need to be in this repository, it can be anywhere. I'll be fine with adding a mention somewhere that there is `pre-commit` config repository(ies) for people who care, but I don't.

I don't want to maintain bits I don't know anything about. I'm sorry, it just doesn't work in the long term (this kind of stuff happens to bitrotten, and I couldn't know, also fixing such unfamiliar stuff is a burden - as a single example is Nix configurations).

PR #710, opened 2024-03-22
cabal 3.10.3

andreasabel:
This completes and updates the `cabalGhcupVersion` map.

Commits:
- **Bump Cabal to latest versions**
- **Update fixtures to cabal 3.10.3.0**


phadej:
This is obsolete, we don't map versions anymore.

PR #715, opened 2024-04-18
Update to v4 actions

phadej:


liskin:
This can't be done while keeping "distribution: bionic" the default...
Otherwise it just starts failing all over the place - https://github.com/xmonad/xmonad/pull/502

phadej:
@liskin good point, will be fixed in https://github.com/haskell-CI/haskell-ci/pull/716

PR #716, opened 2024-04-20
Make Jammy a default distribution

phadej:


liskin:
https://github.com/haskell-CI/haskell-ci/blob/master/cabal.haskell-ci still says "Bionic is a default" and "Supported: xenial, bionic, focal" neither of which is true. 😞

PR #720, opened 2024-05-01
Drop GHC-7 support, Ubuntu Bionic&Xenial; Add Noble Numbat

### LINKED ISSUE #719: What to do with older GHCs.
From https://wiki.ubuntu.com/Releases

Ubuntu Bionic (18.04) is EOL. On Ubuntu Focal (20.04) there are some GHC missing (there is GHC 7.0 and 7.2 in hvr ppa; and GHC-7.10+; in ghcup the GHC-7.10 is IIRC broken (required some post install patching). On Ubuntu Jammy there are only GHC-8.4+ (in `ghcup`, there are no hvr ppa for jammy).

Recently, GHA started warning about newer nodejs, forcing to update the action versions (`checkout`, `cache`). These work on `focal`, but don't on `bionic` (erroring with ``/__e/node20/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)``).

Even if it worked, Bionic is EOL.

So what to do?

A cleanest solution is to drop support for older GHCs. At least GHC-7. I think it's time, I have holding up with these for as long as possible, but now the issue is quite hardly forced.

So, soon, I will remove GHC-7 support from `haskell-ci`.

If you have comments, any good ideas, please comment.

Ping @RyanGlScott, as you are probably the only one running extensive CI setups with GHC-7. I only do because I could, but I'd be fine dropping GHC-7 support as I do sweep across packages updating for GHC-9.10.

--

GHC Version 8.0. 1 (released 21st May 2016), that's *a long time ago*.

phadej:
Resolves #719


liskin:
Is it intentional that haskell-ci silently stops generating CI matrix entries for GHC 7.* without any warning/error whatsoever?

phadej:
> Is it intentional that haskell-ci silently stops generating CI matrix entries for GHC 7.* without any warning/error whatsoever, despite those being mentioned in tested-with?

Unfortunately it's how `haskell-ci` works. We specify a *version range* for GHC versions; if today you write `tested-with: GHC >=10` you won't get none, with `GHC >=7.0` you'll get all (incl all minor versions, i.e. a lot).

OTOH, the 

```
*INFO* Generating GitHub config for testing for GHC versions: 8.0.2 8.2.2 8.4.4 8.6.5 8.8.4 8.10.7 9.0.2 9.2.8 9.4.7 9.6.3 9.8.1
```

output is almos the only thing `haskell-ci` outputs, so should be easy to spot if something as off; and obviously `git diff` before you commit the changes.

So I don't see that "silence" as a major problem.

liskin:
Oh, I didn't realise you can use anything else than `==` in `tested-with` (it doesn't seem to be documented anywhere). Makes sense if you indeed can.

And yeah, as you may have noticed from the GitHub messages about referenced commits/PRs, I have implemented exactly the check you suggest: https://github.com/liskin/xmonad-maintenance/commit/50b9290d8e1a12c9de03b3100e65ef9956c78590

Seemed like a bit of a hack but good to know it's not 🙂

liskin:
It creates a pull request, that's supervised enough for us 😏

PR #733, opened 2024-06-30
Default to cabal-3.12.1.0

andreasabel:
~~Caveat: There is a (imho) serious regression in `cabal check` for 3.12.1.0:~~
- https://github.com/haskell/cabal/issues/10162


phadej:
See #737. The `cabal-install-3.12` has breaking changes, so just bumping version is not enough.

PR #737, opened 2024-07-01
Use cabal-install-3.12.1.0

phadej:
Closes #733


phadej:
This will be delayed by  https://github.com/haskell/cabal/pull/9326 change. I need to update `cabal-install-parsers` and make correspoding fix-release of `cabal-docspec`.

PR #740, opened 2024-07-08
Use ghcup-0.1.30.0

phadej:


src/HaskellCI/Ghcup.hs
```diff
@@ -4,5 +4,5 @@ import HaskellCI.Prelude
 
 import qualified Distribution.Version as C
 
-defaultGhcupVersion :: Version
-defaultGhcupVersion = C.mkVersion [0,1,20,0]
+initialGhcupVersion :: Version
+initialGhcupVersion = C.mkVersion [0,1,20,0]
```
andreasabel, line 8: @phadej Is there a reason why this say "20" instead of "30"?
Is it intentional or was this forgotten to be updated?

phadej: Read the whole diff, incl. `Config.History`. This is the initial, not the current version.

PR #742, opened 2024-07-18
Support for weeder step

### LINKED ISSUE #132: Add support for running weeder
http://hackage.haskell.org/package/weeder is very useful at catching over-specified build-dependencies etc., and it would be nice to have support for running it on travis-ci automatically. Unfortunately, weeder is currently very much tied to `stack`. I suppose that's going to change eventually, though, it's only a matter of time until someone figures out how to support `cabal-install` properly.

kostmo:
Closes #132


phadej (changes requested):
Before I even look at this PR more closely. It contains independent changes like `ghExprWrap` and changes elsewhere. Do them separately. I don't know whether I like them or not.

kostmo:
> independent changes like `ghExprWrap` and changes elsewhere

Ok, I have removed the independent changes from this branch.

src/HaskellCI/GitHub.hs
```diff
@@ -536,6 +536,15 @@ makeGitHub _argv config@Config {..} gitconfig prj jobs@JobVersions {..} = do
             -- disable-documentation disables docs in deps: https://github.com/haskell/cabal/issues/7462
             sh_if range $ "$CABAL v2-haddock --disable-documentation" ++ haddockFlags ++ " $ARG_COMPILER --with-haddock $HADDOCK $ARG_TESTS $ARG_BENCH all"
 
+        when runWeeder $
+            let range = RangeGHC /\ Range cfgWeeder in
+            let ifCond = ghCompilerVersionArithPredicate allVersions range in
+            for_ pkgs $ \Pkg{pkgName} -> do
+                githubUsesIf "weeder" "freckle/weeder-action@v2" ifCond $ buildList $ do
+                    item ("ghc-version", ghWrapExpr "matrix.compilerVersion")
```
phadej, line 544: I don't see this working. Comparing strings like `8.10.7` and `8.8.4` with operators like `>=` wouldn't work as intended.

---

Add test

phadej (changes requested):

src/HaskellCI/GitHub.hs
kostmo: > Comparing strings like `8.10.7` and `8.8.4` with operators like `>=` wouldn't work as intended.

The strings get converted to numeric representation before comparison.  E.g., with [this config](https://github.com/swarm-game/swarm/blob/53b8a3b7ad2a1a874e17fea087cf6ef30957fa53/.github/workflows/cabal.haskell-ci#L16):
```
weeder: >= 9.4 && < 9.8
```
the [resulting comparison](https://github.com/swarm-game/swarm/blob/53b8a3b7ad2a1a874e17fea087cf6ef30957fa53/.github/workflows/haskell-ci.yml#L240) is:
```
${{ env.HCNUMVER >= 90400 && env.HCNUMVER < 90800 }}
```
And here's a [live demonstration](https://github.com/swarm-game/swarm/pull/2060#discussion_r1684957641) featuring coomparison against a number like `9.10.7`.

I have also now added a golden test.

test/Tests.hs
```diff
@@ -37,44 +39,67 @@ main = do
             , fixtureGoldenTest "copy-fields-some"
             , fixtureGoldenTest "copy-fields-none"
             ]
+        , testGroup "github-specific"
```
phadej, line 42: unnecessary complication.

Please, do the least required stuff.

kostmo: Do you mean, remove the extra `testGroup "github-specific"` layer, but keep `fixtureWeederGoldenTest`?  I've removed the extra group layer now.

kostmo: Or if you mean that `fixtureWeederGoldenTest` is too complicated, could you help give an idea of what a minimal test would look like?

PR #745, opened 2024-10-21
Add GHC 9.12.1-alpha1

RyanGlScott:


RyanGlScott:
Sadly, I realized that there is a slight error in this PR right after you merged this: the version number should be `9.12.20241014` instead of `9.12.0.20241014`. (Yes, that's rather odd, but this was how it was done on the [GHC side](https://downloads.haskell.org/~ghc/9.12.1-alpha1/) and [GHCup side](https://github.com/haskell/ghcup-metadata/blob/cc77b39fa92bee71bac3fffb5c784fde9ce805ac/ghcup-prereleases-0.0.8.yaml#L1974).) I'll submit a separate PR to correct this.

andreasabel:
> version number should be `9.12.20241014` instead of `9.12.0.20241014`

No clue why the GHC crowd decided on this ill-logical number (it is BIGGER than 9.12.1 rather than smaller), probably it was an oversight.

phadej:
IT was, see e.g. https://gitlab.haskell.org/ghc/ghc/-/issues/25390

PR #746, opened 2024-10-21
Correct version number for GHC 9.12.1-alpha1

RyanGlScott:
Due to a packaging mistake, GHC 9.12.1-alpha1 uses the version number 9.12.20241014 rather than 9.12.0.20241014. See
https://gitlab.haskell.org/ghc/ghc/-/issues/25123#note_591718.

This updates `haskell-ci` to use the actual version number.

### DIFF
--- a/fixtures/all-versions.github
+++ b/fixtures/all-versions.github
@@ -33,9 +33,9 @@ jobs:
             compilerVersion: "8.4"
             setup-method: hvr-ppa
             allow-failure: false
-          - compiler: ghc-9.12.0.20241014
+          - compiler: ghc-9.12.20241014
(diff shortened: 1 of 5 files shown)


phadej:
This is dubious. `9.12.20241014` is greater then the future real `9.12.1`. I'd rather skip the alpha1 all together. GHC HQ should repackage.

RyanGlScott:
OK. In that case, can you revert the changes from https://github.com/haskell-CI/haskell-ci/pull/745? They're functionally broken.

phadej:
Reverted and opened https://gitlab.haskell.org/ghc/ghc/-/issues/25408

PR #753, opened 2024-11-19
Fix suspected typo in Config/History.hs

andreasabel:
I suppose 2024114 was meant to be 20241114 in the version history.


phadej:
fixed as part of https://github.com/haskell-CI/haskell-ci/pull/754

PR #759, opened 2024-12-19
GHC-9.12.1 proper

phadej:


src/HaskellCI/Config/History.hs
```diff
@@ -54,6 +54,14 @@ configHistory =
         & field @"cfgVersionMapping" .~ Map.singleton (mkVersion [9,12,1]) (mkVersion [9,12,0,20241128])
     , ver 0 19 20241218 := \cfg -> cfg
         & field @"cfgCabalInstallVersion" ?~ C.mkVersion [3,15,0,0,2024,10,3]
+    , ver 0 19 20241202 := \cfg -> cfg
```
andreasabel, line 57: There are now two entries for version `0.19.20241202`.  Is this intended?  How is this to be understood?
Maybe this is a typo and you wanted something like `20241219`?

phadej: Yes, it's a typo; but nothing will be broken so far (the version here is not used for anything, and thus not tested).

PR #797, opened 2025-09-08
v0.19.20250908: Github target: Use actions/checkout v5

andreasabel:
Dependabot complains about `actions/checkout@v4`


andreasabel:
The testsuite golden values need to be updated still.

(However, given past experience of contributing to this repo, I won't put in more work until I get the approval that this will get merged.)

